PR #58, opened 2014-12-15
Enriched entity rendering

EtlamGit:
The Entity rendering introduced with the overlay feature lacks good Entity categorization. E.g. ArmorStands are handled as hostile Mobs. In my opinion it is impossible to detect the category of an Entity automatically and 100% correct; especially after Minecraft updates or for mod packs. So I decided to implement an EntityIdentifier to manage definitions for known Entities. It works similar to the other Identifiers and with a JSON configuration file. For future updates only that definition file has to be updated, not the code. With this Identifier it was also simple to add individual rendering for each type of Entity. I used a color code naturally derived from the Entity (e.g. a Pig is pink and a neutral Mob, Blazes are yellow and hostile, ...)
* implemented EntityIdentifier
* using EntityIdentifier to categorize and render Entities
* the Overlay menus are reduced by one hierarchy to reduce clicking
* Entities too deep under ground are hidden 
* there is a readme_entity.txt describing the syntax
* some code refactoring to get consistent naming

Somewhere in the last month some kind of bug was introduced: I get an access exception at the moment the application quits. Up to now I was not able to find that bug nor the commit it was introduced. But it was probably in the last 3 month, as I did not observe that before. It is already in the master branch and not introduced in this PullRequest.


EtlamGit:
@mrkite: when you are testing this code, you will notice usage of a really nice Qt function QColor::setNamedColor. It allows to use descriptive textual color names in the configuration files. At the moment it is only used for the EntityIdentifer while parsing vanilla_entity.json, but it could also be used for all the other colors defined in JSON files. The only thing to be changed would be to add a "#" in front of each hex color in the current JSON files.

Some PRO arguments:
+ descriptive names possible
+ much more other color description syntax possible
+ same syntax for all configuration files
Some CON arguments:
- not explicitly needed for blocks, as their color is a derived hex value
- small change to the syntax of vanilla_blocks.json

If you think it would be a nice change, I would volunteer to implement that changes to vanilla_blocks.json and BlockIdentifier.

EtlamGit:
@mrkite Hi, any thoughts about my last comment?
Or suggestions what I should improve to get the PullRequest accepted?

mrkite:
I'm not against changing the syntax of the jsons.  I'd also like to move the "official" default jsons to github.

Issue #60, opened 2014-12-24
Saving PNG - Minutor.exe stopped working


eyekanspele:
When trying to load my vanilla SMP world files in minutor works no problem as expected. However when I go to save the world as a PNG; it lets me chose a directory and file to save the png, however when I go to save right away I am given a windows error screen that "Minutor has stopped working".

This is happening to me with both the installed and non-installed versions of version 2.0.1.

I am using a x64 version of Windows 8.1 with an intel based cpu.

george-tsiros:
getting what i think is a null deref when i click the menu item. windows 10, amd64, latest minutor release from github releases (5 days ago)?

Exception thrown at 0x00007FF7636915E0 in minutor.exe: 0xC0000005: Access violation reading location 0x0000000000000000.

which of course means little, there's no debugging information as far as i can see and i fear building it requires qt which is a massive headache in and of itself (did you really have to use qt...?) . The entirety of the world folder is 128MiB. It's a 1.17.1 world. 

strangely enough, when i try to save a _snapshot_ world, it works ok?

EtlamGit:
PNG save "should" work with Minecraft 1.17.1 (I am using it).

Please [try to find a way](https://www.google.com/search?client=firefox-b-d&q=share+big+files+without+registration) to upload your world so we can start debugging.

george-tsiros:
torrent ok?

[New_World.zip](https://github.com/mrkite/minutor/files/7448275/New_World.zip)

The .7z inside is the entirety of the folder. For your convenience, either download only the .7z or only the rest of the files.

I am sorry I could not find a way to do a magnet link. In the meantime i will try with a new blank 1.17.1 world see what it does

george-tsiros:
DON'T bother with THAT torrent, it seems a completely new 1.17.1 world _also_ fails. So it is not something with the world, should be something with my pc. Please contact me somehow.

EtlamGit:
I noticed some empty region files in your world. This was a bug with 1.17.

There is already a patch for this issue: de9c6377f214f32a121356eb1442ce48910aa196 that will be released together with the 1.18 update.

You can test that with a [development preview build](https://www.dropbox.com/s/c9n320hjet3hdjf/minutor.zip?dl=0).

george-tsiros:
ooooh!

you're good!

PR #62, opened 2015-01-20
fix worldsave when backup files are present

EtlamGit:
When backup files from other tools are present in the region folder
Minutor could hang infinitely while trying to calculate the world
border. This commit completely rewrites the Chunk number from region
file detection code.
This PullRequest closes issue #60

### DIFF
--- a/worldsave.cpp
+++ b/worldsave.cpp
@@ -224,27 +224,39 @@ void WorldSave::findBounds(QString path, int *top, int *left, int *bottom, int *
 	//loop through all region files and find the extremes
 	while (it.hasNext())
 	{
-		QString fn=it.next();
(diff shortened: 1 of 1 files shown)


EtlamGit:
@mrkite : I know that this change is not necessary for normal operation, but it makes Minutor more robust to leftovers from other tools.
Just write a short comment if I should change something or you dislike parts of my changes.

EtlamGit:
Hi @mrkite are you still doing Minecraft related stuff?
I personally reduced it much in the last weeks... but I still use Minutor.

mrkite:
I'm not playing Minecraft anymore, but every few months I make sure Minutor is still working.  Sorry about taking so long to merge this.

PR #66, opened 2015-05-30
Mooshrooms are not shown on the map

Metibor:
The savegame ID for Mooshrooms is MushroomCow


EtlamGit:
You are right.
And the same is true for Minecart -> MinecartRidable
MinecartSpawner and EnderCrystal are missing (LeashKnot also missing, but is not stored on disk).

I will add those in another PullRequest...

mrkite:
Merged via the other PR.

PR #74, opened 2015-11-01
added Region & Chunk checkerboard

EtlamGit:
- added command line parameters to write the PNG image with a Region and/or Chunk checkerboard pattern
- enhanced command line long parameter parsing: mixed lower and higher case are allowed now

PR #75, opened 2015-11-09
fixed bug with Red Sandstone Slab & Stairs

EtlamGit:
Mob spawn detection was not working on Red Sandstone Slab & Stairs
Added "Upper" variant name for all Slabs.

### DIFF
--- a/definitions/vanilla_ids.json
+++ b/definitions/vanilla_ids.json
@@ -611,7 +611,6 @@
       "id": 44,
       "name": "Stone Slab",
       "color": "a3a3a3",
-      "mask": 7,
       "transparent": true,
       "variants": [
         {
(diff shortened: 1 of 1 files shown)


EtlamGit:
Hello @mrkite , could you please merge this PullRequest?
I have prepared and double checked all 1.9 changes, but I would like to generate that new PullRequest as an increment on the current changes (maybe also #74 #72).

mrkite:
In your next PR, with the 1.9 changes, were you going to go ahead and change the update urls on the definitions to point directly to github?  (Which I still think is an excellent idea).  I can do it, but if you already did it, I don't want any merge conflicts.

EtlamGit:
The plan is to have the URLs from GitHub. I will double check, that they are in.

PR #77, opened 2016-03-18
Definition update 1 9

EtlamGit:
All BLOCK updates and the new Shulker from Minecraft 1.9.
As additional bonus the definition file updates are now fetched directly from GitHub.

(Something got puzzled in my branch, but I hope everything is sorted the right way now...)

### DIFF
--- a/definitions/vanilla_biomes.json
+++ b/definitions/vanilla_biomes.json
@@ -248,5 +248,5 @@
       "name": "Mesa Plateau M"
     }
   ],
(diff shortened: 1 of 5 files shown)


mrkite:
Now I really need to release new binaries.  Thanks for all your hard work.

zeroeth:
+1 for binaries! some of us are trapped in remote regions of the world after nether portals gone wrong :)

PR #79, opened 2016-04-24
Feature single layer mode

hidroto:
This is a simple patch to add the feature in issue #6
it does not fulfill all of the requirements but should make a good base to expand upon.

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -110,6 +110,11 @@ int main(int argc,char *argv[])
 			minutor.setViewDepthshading(true);
 			continue;
 		}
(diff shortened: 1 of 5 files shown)


EtlamGit:
Hello and sorry, that nobody found the time to give you some feedback...

This feature was also on my (long) list with things to be implemented some time. The changed files look similar to what I would have done, but I did not test your code jet.
Unfortunately your code is not mergable at the moment, because the complete code refactoring to the Google C++ code style.

I see two possible options for you:
- update your branch to the latest changes
- wait until I have implemented a similar solution

EtlamGit:
@hidroto additional comments from your side?

PR #81, opened 2016-05-24
fix Entity rendering depth

EtlamGit:
Now render correctly Entities inside the current depth layer. Before this change, some Mobs may be invisible when the depth slider is at the same level.
(also done some TAB vs. spaces fixes)


mrkite:
I'm going to do a complete linter pass and standardize the code format one of these days.

PR #82, opened 2016-05-24
added detection for EndCity structures

EtlamGit:
All new structures have to be linked to the correct dimension (if not Overworld).

### DIFF
--- a/generatedstructure.cpp
+++ b/generatedstructure.cpp
@@ -51,6 +51,10 @@ QList<QSharedPointer<GeneratedStructure> > GeneratedStructure::tryParse(Tag* dat
 								{
 									structure->setDimension("nether");
 								}
+								else if (structure->type() == "Structure.EndCity")
+								{
+									structure->setDimension("end");
+								}
 								else
 								{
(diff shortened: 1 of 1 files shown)


mrkite:
Sorry I didn't pull this because I was in the middle of reformatting all the code.

EtlamGit:
Completely unrelated to this pull request...

I just noticed that you updated the starting page and at first describe the web based pack builder. As I never took a look to that code, I assume it is not working with the latest changes to the definition file syntax. The files you provide at your website are also not here in the repository. Probably they have to be looked trough and adapted slightly.
Did you test them with Minutor 2.1?

mrkite:
Yeah, the code needs to be updated... which honestly isn't a lot of work.  I'll do it this week.

I'm wondering, however, if it wouldn't be better in the future to build a pack editor directly into Minutor.  That's a bit more work though.

EtlamGit:
You are right with "isn't a lot work" when you refer to blocks and block colors.
But it took me several month of external code analysis to get the Mob spawn detection work the correct way.

Issue #83, opened 2016-06-11
Version 2.1 Save PNG doesn't append file type


BigAlanM:
I have just started using the latest release of Minutor 2.1.0 and tried to save a PNG file.
The file was created without ".png" in it's name.

mrkite:
Fixed by Etlam.

Issue #34, opened 2014-02-17
missing suffix ".png" for world save


EtlamGit:
When writing the world into a PNG image, no default suffix is set. If the given filename contains no suffix, the resulting image has none.

PR #86, opened 2016-06-12
fix for missing .png suffix

EtlamGit:
This fixes issue #83. Intentionally it should already have been fixed with commit 87c9af5, as this problem is a duplicate of #34. 

The problem seems to be Operation System or Qt version dependand, therefore I added much more check code to handle this problem in any case


mrkite:
Qt sometimes drives me crazy with it's OS-dependent functionality.

PR #89, opened 2016-06-15
Update 1.10

EtlamGit:
16w20a:
add Entity: Polar Bear
add Block: Magma Block
add Block: Nether Wart Block
add Block: Red Nether Brick
add Block: Bone Block

Also Entities can have individual names now. This would enable translated definition files.


mrkite:
i18n for definition files is an interesting problem.

PR #96, opened 2016-10-21
Biome colors

highthunder:
Add a viewing mode in which the color of a block is determined by the biome it is in. It is very similar to Amidst but it is capable of viewing worlds in which terrain was generated across several versions.


IronException:
I would also be cool to just see the biome color

PR #98, opened 2016-11-18
Update 1.11

EtlamGit:
**16w32a:**
Entity names were completely renamed by Mojang. Therefore some backwards compatible code and definition was implemented. Lower case version are handled automatically, but renamed Entities have to be specified with "id1" tag.
* Horse split into entity IDs horse, donkey, mule, skeleton_horse and zombie_horse
* Skeleton split into entity IDs skeleton, stray and wither_skeleton
* Zombie split into entity IDs zombie, zombie_villager and husk
* Guardian split into entity IDs guardian and elder_guardian

**16w39a:**
+ Block: Observer
+ Block: (16 colored) Shulker Boxe(s)
+ Entity: Llama
+ Entity: Vindicator
+ Entity: Evoker
+ Entity: Vex

### DIFF
--- a/definitions/readme_entity.txt
+++ b/definitions/readme_entity.txt
@@ -30,6 +30,10 @@ the category if omitted. The highlight color (unique for each entity) is given
 by "color". If omitted a pseudo random color is calculated based on the hashed id.
 The "name" is automatically copied from the "id", but can also be set individually.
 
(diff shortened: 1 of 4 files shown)


EtlamGit:
Just as as additional comment: Old versions of Minutor will work with the new definition files!

Issue #103, opened 2016-12-03
Definition Update not working


EtlamGit:
Triggered by issue #102 I tested functionality of definition update: It fails silently

EtlamGit:
During debugging I get a `QNetworkReply::ProtocolUnknownError` in definitionupdater.cpp::checkTime()

EtlamGit:
It seems that for HTTPS it is necessary to self-compile Qt with OpenSSL support linked to the library:
https://forum.qt.io/topic/44117/qnetworkaccessmanager-and-https/18
http://doc.qt.io/qt-5/ssl.html

upwhere:
EtlamGit writes:
> It seems that for HTTPS it is necessary to self-compile Qt
> with OpenSSL support linked to the library:

I don't think we want to bundle Qt so that isn't really an
option.

Are we left with:
- Find a cross-platform https client
- Mirror the files on another protocol (http, ftp)
Or do we have more options?

seancode's website works with http but I don't know if a mirror
can be set up there and whether or not the host wants to.

I could offer a mirror myself but then everyone would have to
trust in me to stay available and not serve anything malicious.

madmaxoft:
If it's just definitions, and not a program file, then plain http is okay, I guess, there's very little room for anything malicious. If it were for the main program files, though, that's a different story.

Anyway, I do have Qt self-compiled because of Minutor already, because I wanted to link Qt statically, rather than dynamically. I could have a look at enabling OpenSSL in my build and then provide the executables for the releases.

mrkite:
Yeah, I already statically link Qt for Windows and OSX releases.  Linux releases on the other hand are dynamically linked (package manager for the win).

I'd like to keep the definitions on github, because it's just easier for everyone... so getting https working is definitely the way to go.

EtlamGit:
Github only supports HTTPS. So moving the definition files there caused this problem.

I'm able to build OpenSSL from the sources. Versions > 1.0.0 are claimed to be compatible with Qt. But versions 1.1.x changed a lot and I'm not able to build it anymore. Pre-build DLLs are also available for download from openssl.org.

From what I read during the day is that Qt "should" load provided OpenSSL DLLs automatically. But with my static build it does not . . .
It seems that without specifying it, configuration defaults to -no-openssl whereas automatic DLL loading would be -openssl (next Qt build should be finished tomorrow...)

[...]

PR #104, opened 2016-12-10
Fix automatic definition file update via https

EtlamGit:
for HTTPS support a recompile of Qt is necessary (HTTP would work without)

github.com does not provide a LastModification HTTP header response. Therefore the completely changed update code gets the remote definition file, extracts the version string and compares it with the local version.

Automatic update is done only once a week.

This should fix issue #103 at least it does it during my debugging sessions.

(contains also some name refactoring to align naming of objects through the code)

### DIFF
--- a/definitionmanager.cpp
+++ b/definitionmanager.cpp
@@ -20,7 +20,8 @@
 #include "./zipreader.h"
 #include "./definitionupdater.h"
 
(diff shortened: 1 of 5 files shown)


mrkite:
I'm recompiling Qt right now.

Issue #111, opened 2017-03-28
Can't build


ParkerMc:
I can't build it from git I am using QtCreator on Arch Linux everything is up to date. Here is the build output. note: I replace my system username with `****`
```
15:03:17: Running steps for project minutor...
15:03:17: Starting: "/usr/sbin/qmake" /home/****/git/minutor/minutor.pro -spec linux-g++ CONFIG+=debug CONFIG+=qml_debug
Info: creating stash file /home/****/git/build-minutor-Desktop-Debug/.qmake.stash
15:03:17: The process "/usr/sbin/qmake" exited normally.
15:03:17: Starting: "/usr/bin/make" qmake_all
make: Nothing to be done for 'qmake_all'.
15:03:18: The process "/usr/bin/make" exited normally.
15:03:18: Starting: "/usr/bin/make" 
/usr/bin/uic ../minutor/properties.ui -o ui_properties.h
/usr/bin/uic ../minutor/settings.ui -o ui_settings.h
g++ -c -pipe -g -std=gnu++11 -Wall -W -D_REENTRANT -fPIC -DQT_QML_DEBUG -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -I../minutor -I. -I../minutor -isystem /usr/include/qt -isystem /usr/include/qt/QtWidgets -isystem /usr/include/qt/QtGui -isystem /usr/include/qt/QtNetwork -isystem /usr/include/qt/QtCore -I. -isystem /usr/include/libdrm -I. -I/usr/lib/qt/mkspecs/linux-g++ -o labelledslider.o ../minutor/labelledslider.cpp
g++ -c -pipe -g -std=gnu++11 -Wall -W -D_REENTRANT -fPIC -DQT_QML_DEBUG -DQT_WIDGETS_LIB -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_CORE_LIB -I../minutor -I. -I../minutor -isystem /usr/include/qt -isystem /usr/include/qt/QtWidgets -isystem /usr/include/qt/QtGui -isystem /usr/include/qt/QtNetwork -isystem /usr/include/qt/QtCore -I. -isystem /usr/include/libdrm -I. -I/usr/lib/qt/mkspecs/linux-g++ -o biomeidentifier.o ../minutor/biomeidentifier.cpp
[...]

PR #112, opened 2017-03-29
Fixed compiler errors

ParkerMc:
This is the fix for #111

### DIFF
--- a/biomeidentifier.cpp
+++ b/biomeidentifier.cpp
@@ -89,7 +89,7 @@ QColor BiomeInfo::getBiomeColor( float temperature, float humidity, int elevatio
   int g = (int)std::clamp( green, 0.0f, 255.0f );
   int b = (int)std::clamp( blue,  0.0f, 255.0f );
 
-  return QColor(r,g,b);
+  return QColor::fromRgb(r,g,b);
 }
 
 QColor BiomeInfo::mixColor( QColor colorizer, QColor blockcolor )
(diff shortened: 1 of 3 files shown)


biomeidentifier.cpp
```diff
@@ -113,7 +114,7 @@ QColor BiomeInfo::mixColor( QColor colorizer, QColor blockcolor )
   }
 //float lig = std::clamp( ligC + ligB - 0.5f, 0.0f, 1.0f );  // more brightness contrast
   float lig = (ligC + ligB ) / 2.0f;
-  return QColor::QColor().fromHslF( hue, sat, lig );
+  return QColor().fromHslF( hue, sat, lig );
```
madmaxoft: The `fromHslf` function is static, so why not use the simple `QColor::fromHslf(...)`?

biomeidentifier.cpp
```diff
@@ -127,7 +128,7 @@ QColor BiomeInfo::getBiomeGrassColor( QColor blockcolor, int elevation )
     // perlin noise generator omitted due to performance reasons
     // otherwise the random temperature distribution selects
     // (below -0.1°C) ‭4C.76.3C‬ or ‭6A.70.39 (above -0.1°C)
-    colorizer = QColor::QColor(0x6a,0x70,0x39);  // hard wired
+    colorizer = QColor(0x6a,0x70,0x39);  // hard wired
```
madmaxoft: Consider using `QColor::fromRgb(...)` to explicitly show what color system is used. This goes for the other similar changes, too.

ParkerMc:
I think that is all the changes you wanted.

Issue #109, opened 2017-03-04
Feature: Use accellerator keys for variable scrolling speed


Metibor:
When using the arrow keys for scrolling, it would be nice if you can use some accellerator keys to change the size of the steps (and be faster than dragging the map with the mouse).
Currently the step size is 10 blocks.

Suggestion:
Unmodified: Change from 10 to 16 blocks per step (1 Chunk)
SHIFT: 1 block per step for fine positioning (yes, I need it sometimes)
CTRL: 128 blocks per step (8 Chunks)
ALT: 512 blocks per step (1 Region)

mrkite:
Huh.  I could've sworn it did this already, but apparently not.  It's simple enough to add, so I'll add it.

PR #114, opened 2017-04-11
Removed zoom factor from keyboard movement and did some code clean-up (#109)

Metibor:
I think the zoom factor is not needed.
Also the type of stepSize was wrong before. Without the factor an int should be correct.

### DIFF
--- a/mapview.cpp
+++ b/mapview.cpp
@@ -168,40 +168,40 @@ void MapView::wheelEvent(QWheelEvent *event) {
 
 void MapView::keyPressEvent(QKeyEvent *event) {
   // default: 16 blocks / 1 chunk
-  int stepSize = 16;
+  float stepSize = 16.0;
 
   if ((event->modifiers() & Qt::ShiftModifier) == Qt::ShiftModifier) {
     // 1 block for fine tuning
-    stepSize = 1;
+    stepSize = 1.0;
   }
   else if ((event->modifiers() & Qt::AltModifier) == Qt::AltModifier) {
     // 8 chunks
-    stepSize = 128;
+    stepSize = 128.0;
     if ((event->modifiers() & Qt::ControlModifier) == Qt::ControlModifier) {
       // 32 chunks / 1 Region
-      stepSize = 512;
+      stepSize = 512.0;
     }
   }
 
   switch (event->key()) {
     case Qt::Key_Up:
     case Qt::Key_W:
-      z -= stepSize;
+      z -= stepSize / zoom;
       redraw();
       break;
     case Qt::Key_Down:
     case Qt::Key_S:
-      z += stepSize;
+      z += stepSize / zoom;
       redraw();
       break;
     case Qt::Key_Left:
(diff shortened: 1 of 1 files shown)


mrkite:
I'm not going to merge this just yet because I need to test something.  The zoom factor is there so that the map moves the same amount regardless of zoom.  If you zoom fully in, pressing the left arrow will scroll the map by (completely made up number) 30 pixels, when you zoom fully out, it still moves the map by 30 pixels.

If you zoom completely out and use the arrow keys, the map would pan excrutiatingly slow otherwise.

Metibor:
By removing the factor, I wanted to switch from pixel based movement to block based movement.

**Current situation (stepSize / zoom)**
Normal zoom (1 block is displayed as 1 pixel):
If I press the an arrow key once, the map moves 16 pixels which is 16 blocks.
Zoomed in (i.e. 1 block is displayed as 4 pixels):
If I press the an arrow key once, the map moves 16 pixels which is only 4 blocks.

=> So the movement is pixel-based

**Proposed (just stepSize)**
Normal zoom (1 block is displayed as 1 pixel):
If I press the an arrow key once, the map moves 16 pixels which is 16 blocks.
Zoomed in (i.e. 1 block is displayed as 4 pixels):
If I press the an arrow key once, the map moves 64 pixels which is still 16 blocks.

=> Now the movement is block-based

I think that this is more intuitive, but I have to admit the following:
If you zoom close to the map (i.e. factor 16) and make bigger steps (i.e. Alt+Arrow) the movement is really big (16 * 128) and you can't see where you are going anymore.

Maybe I can make it an option in the settings screen and use the old as default...

EtlamGit:
What about:
- moving pixel based and with zoom factor when no modifier keys are pressed
- move block/chunk/region based when the corresponding modifier key is pressed

I think it is also useful, to get the left-top pixel aligned to the block/chunk/region I am moving in that moment. Meaning, no half chunks at left and top border.

Metibor:
Hi, I readded the zoom factor, so the only change in this pull request is changing my int to float.

The half-pixel-problem exists in both versions, but with the pixel based movement (with the factor) it is easier to remove them (Shift+Arrow).

Metibor:
Oh, you said half-chunks. You can't see them currently, because there is no real indicator for that. Maybe we could add an overlay for chunk/region borders?

EtlamGit:
We have an overlay for them already, but also only via [command line](https://github.com/mrkite/minutor/wiki/Command-Line) ...
(I use them during automatic map export.)

My suggestion was just to align the left-topmost pixel of the map to a chunk/region when in special keyboard navigation mode. That would make it easier to do screen shots or plan chunk-border dependent builds.
But you are right, adding an control element in the GUI to toggle this feature is something I could do also.

Metibor:
Please, could you merge my changes from April 14th, or are there some problems left?
The current navigation is based on the center pixel. Changing that, should be another issue.

mrkite:
Yeah, sorry about that.

PR #117, opened 2017-06-17
Update 1.12

EtlamGit:
We should do a new release, as the automatic definition update is not working for 2.1.0 from last year.
I plan to do color palette adaption to latest Mojang changes and wiki details next month.


mrkite:
I'll try to put together a new binary, and make sure https support works properly.

EtlamGit:
During July I have some spare time and plan to extend the Wiki with more compile instructions.
In case you encounter problems we could try to identify all tiny steps towards a binary.
But I have not build any Mac or Linux binary!

mrkite:
The linux binaries are pretty well spelled out in the readme.  The mac binaries are very similar to the Windows ones (having to create a static library first).

mrkite:
I updated the binaries.  Haven't checked to see if the static library works properly with SSL yet though.

PR #118, opened 2017-07-12
Some naming corrections and small additions

Metibor:
I made the following changes:
The entity ID is changed from Ozelot to ocelot in 16w32a
<color> Stained Terracotta is only called <color> Terracotta
Structure Block are now more precise. Colors are based on the average value from the texture
For completeness I added the Structure Void block

Some small questions:
How is your method for defining the colors in the definition file?
Are you planing to reflect the color changes for wool, stayned glass, shulker boxes, etc in 1.12?
Is it possible to add the new colors for the bed?

### DIFF
--- a/definitions/vanilla_entity.json
+++ b/definitions/vanilla_entity.json
@@ -181,6 +181,8 @@
         },
         {
           "id": "Ozelot",
+          "id1": "ocelot",
+          "name": "Ocelot",
           "color": "#808000"
         },
         {
--- a/definitions/vanilla_ids.json
+++ b/definitions/vanilla_ids.json
@@ -1848,82 +1848,82 @@
     },
     {
       "id": 159,
-      "name": "White Stained Terracotta",
+      "name": "White Terracotta",
       "color": "#d1b1a1",
       "variants": [
         {
           "data": 1,
-          "name": "Orange Stained Terracotta",
+          "name": "Orange Terracotta",
           "color": "#a55728"
         },
         {
           "data": 2,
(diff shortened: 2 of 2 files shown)


mrkite:
I did the original colors by shrinking the textures down to a single pixel and using that final averaged color, except for a few blocks where I just picked a representative color, like the red for a flower, etc.

It'd be nice to have support for all variations of blocks.

EtlamGit:
You are right with the oZelot, it did not only change to lower case, it is now also spelled differently.

The Minecraft Wiki still mentions Stained Terracotta, but the official names seem to be without 'Stained'.

I assume these structure blocks will never occur on a saved map. And in case they do, the default light purple is a good color for them. But as you added them it is also fine.

I use average color for all new blocks I define, scaling a block texture to 1x1 is one way to do that. But you have to take care which interpolation filter is used, as this may generate different results. Also for semi transparent blocks it is likely to generate much too dark average color. It is quite annoying to do all this scaling and hex formatting for the new blocks. If you have a nice script, e.g. with ImageMagic I would appreciate to get it...

Updating all the changed colors (Wool etc) was on my summer vacation list, but I did not find any time at all :-( Maybe during Christmas...

It is not (easy) possible to add colored beds, as this information is not stored in the block data as for other colored blocks. It is stored as [entity metadata](https://minecraft.gamepedia.com/Chunk_format#Block_entity_format) like the content of chests. To get the colors you would have to parse and link these database also. In my opinion it is not worth the effort. We would also need some caching and filtering of all these additional data to fit into memory (only loading visible entities).

PR #121, opened 2017-10-09
Settings: Fixed loading and defaults.

### LINKED ISSUE #93: Cannot open saved worlds from the File - Open World menu on Mac
OS X 10.11.6, Minutor v2.1.0, Minecraft 1.10.2

Worlds are saved in the default location: /Users/<username>/Library/Application Support/minecraft/saves

There are several saved worlds therein.

Starting Minutor and selecting File - Open World from the menu does not show any worlds to open.

It is still possible to select File - Open and manually browse to the world's save directory to open a world, but as the Library folder is hidden this is an extra complication.

madmaxoft:
The "use default location" was not applied when starting Minutor. This change produces more understandable flow of the settings initialization.

Fixes #93, at least on Windows.

### DIFF
--- a/settings.h
+++ b/settings.h
@@ -15,6 +15,10 @@ class Settings : public QDialog {
   bool verticalDepth;
   QString mcpath;
 
+
+  /** Returns the default path to be used for Minecraft location. */
+  static QString getDefaultLocation();
+
  signals:
   void settingsUpdated();
   void locationChanged(const QString &loc);
--- a/settings.cpp
+++ b/settings.cpp
@@ -30,17 +30,39 @@ Settings::Settings(QWidget *parent) : QDialog(parent) {
   connect(m_ui.checkBox_AutoUpdate, SIGNAL(toggled(bool)),
           this, SLOT(toggleAutoUpdate(bool)));
 
-  // set stored settings
+  // Load the settings:
   QSettings info;
-  m_ui.checkBox_AutoUpdate->setChecked(info.value("autoupdate",
-                                                  true).toBool());
-  m_ui.lineEdit_Location->setText(info.value("mcdir", "").toString());
-  m_ui.lineEdit_Location->setDisabled(
-      m_ui.checkBox_DefaultLocation->isChecked());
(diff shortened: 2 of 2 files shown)


EtlamGit:
Hi @madmaxoft , as it seems that there are some ambiguities what location to choose in the settings dialog, I see these options for us to improve:
* write an extra comment to use the Minecraft base folder not the saves folder
* automatically detect a wrong chosen saves folder and remove it
* work with both variants, as the saves folder may be named differently anyway

While thinking more about it, I prefer the last option. What do you think?

madmaxoft:
I'm always for making things easier for the end user, so working with both would be my choice. However, it is outside of the scope of this PR, I think.

madmaxoft:
@EtlamGit How about this behavior:
- The Settings object always contains the base folder
- When the user browses for the "Minecraft Location", the code in the Browse button handler inspects the folder they choose. If it is named "saves" and has no "saves" subfolder in it, the code uses the parent folder silently.

This should work for all the cases and prevent any confusion. The only case it wouldn't work would be if the user had a world named "saves" and they chose the saves folder instead of the base folder. But I think it's better to leave it like that.

EtlamGit:
(The only save method would be to search for a folder that contains subfolders which themselves contain a file called level.dat that parses to valid NBT data.)

I like your approach...

mrkite:
My only concern is you don't want the program walking all over the place, recursively checking directories for level.dat.  Especially if they point to something really slow like a network drive.  So we definitely want to keep the automatic checking to a minimum.

settings.h
```diff
@@ -15,6 +15,10 @@ class Settings : public QDialog {
   bool verticalDepth;
   QString mcpath;
 
+
+  /** Returns the default path to be used for Minecraft location. */
+  static QString getDefaultLocation();
```
tigerw: @madmaxoft xoft, is this by any chance, a Cuberite-style comment, in *minutor*?

tigerw: What is this, some sort of crossover episode?

Issue #27, opened 2013-12-02
"not a valid Win32 application"


froggyt:
I use Windows XP Pro. v2002, SP3 and have only been able to run Minutor v1.6.3

Version 2.0.1, regardless of how I install it (installer or zipped minutor.exe), does not run on my system. A message box pops up with the message "C:\program files\minutor\minutor.exe is not a valid Win32 application." 

minutor.exe (v2.0.1) MD5 hash value: 87334d7d8abd6a483ed5ea2beecbc4ed

One other unidentified v2.x release previous to v2.0.1 also has the same problem.

nerijus:
I found that:
Visual Studio 2012 Update 1 has now been released, and adds official support for running apps built with VC++ 2012 on Windows XP.

PR #123, opened 2017-11-19
Enable compilation on non-Linux unix such as FreeBSD

distalzou:
Only use _SC_AVPHYS_PAGES on systems where it exists.  Otherwise fall
back to the default of 10000 chunks.


EtlamGit:
Looks ok to me. And line #27 was a duplicate anyway.

Issue #125, opened 2018-03-09
Program does not open development beta Minecraft 18x10b


erich666:
The latest Minecraft beta, 18x10b, makes world files that cannot be opened by Minutor, which crashes. I suspect 18x10a, https://minecraft.gamepedia.com/18w10a, has the same problem, and my guess is that restructuring for the coral plants is causing problems. Given that Minecraft "used up" all the basic block IDs 0-255 (except for 253 and 254), https://minecraft.gamepedia.com/Java_Edition_data_values#Block_IDs, I suspect they're now using some new extension method that breaks readers - that'll be fun for everyone.

I hope to look into this problem in a bit (Mineways is built off an old version of minutor), but thought I'd give a heads-up and see if anyone's figured it out already.

mrkite:
I haven't been keeping up with Minecraft, so this is definitely news to me.  I'll have to find some time to install Minecraft and see where things go haywire.

EtlamGit:
As far as I know, the next major release from Minecraft may bring a completely new storage format, but development is not finished now. I remember reading something like: worlds are converted on first load, old structures will not be converted properly, no way back, ...

erich666:
I finally had some time to look at this today (after recovering from updating to VS 2017). Yes, they're having fun with the file format. For starters, Biomes went from 256 bytes to 256 integers, and it looks like the fourth (high) byte of the integer is the biome value.

New format is on the left, as seen in NBT Explorer:

![image](https://user-images.githubusercontent.com/803995/37364292-79548a30-26d0-11e8-9aca-a0fd3186c1be.png)

Sections are now pretty interesting, with each 256 slice having a palette associated with it. BlockStates are now 256 long integers. Basically, the byte per block assumption is now gone, is my guess. Anyway, that's as far as I got. Good times.

nnooney:
I've updated Minutor to parse the new format described above. Thanks @erich666 for providing the format that allowed me to understand the new format. The pull request I referenced is very rough around the edges, but I am able to open worlds that I generated in the 1.13 snapshots.

More work needs to be done to update the Definition Manager with the new block format, and there's some room for code cleanup. However, it should be enough to load worlds in Minecraft 1.13+ (I certainly broke backwards compatibility).

erich666:
And just in case, Minecraft 1.13 is now officially out, as of yesterday https://minecraft.gamepedia.com/1.13

EtlamGit:
[...]

PR #127, opened 2018-07-18
Initial 1.13 update

nnooney:
This branch prepares Minutor for parsing files generated in Minecraft 1.13. Notably, it contains the following changes:
- Addition of New NBT Tag "Tag_Long_Array"
- Updates to chunk.cpp to parse new chunk format
- Begin flattening of vanilla_ids.json
- Update mapview.cpp to remove "data" tag
- Update blockidentifier to search based upon block name

I've tested these changes on my world which was generated in 1.13-pre10. The vanilla_ids.json still needs more work, as well as the mapview and blockidentifier, but this should be a start for those who want to look at their maps in the new 1.13 format.

### DIFF
--- a/blockidentifier.cpp
+++ b/blockidentifier.cpp
@@ -53,7 +53,7 @@ void BlockInfo::setName(const QString & newname) {
       !this->name.contains("Full");
   snow = this->name.contains("Snow");
   // precompute biome based watercolormodifier
-  water = this->name.contains("Water");
+  water = this->name.contains("water");
 }
 
 const QString & BlockInfo::getName() { return name; }
--- a/blockidentifier.cpp
+++ b/blockidentifier.cpp
@@ -124,6 +124,7 @@ BlockInfo &BlockIdentifier::getBlock(QString name, int data) {
   //   }
   // }
   // no blocks at all found.. dammit
+  unknownBlock.setName(name);
   return unknownBlock;
 }
 
--- a/blockidentifier.cpp
+++ b/blockidentifier.cpp
@@ -90,37 +90,39 @@ BlockIdentifier::~BlockIdentifier() {
 }
 
 // this routine is ridiculously slow
-BlockInfo &BlockIdentifier::getBlock(int id, int data) {
+BlockInfo &BlockIdentifier::getBlock(QString name, int data) {
   // first apply the mask
-  if (blocks.contains(id))
-    data &= blocks[id].first()->mask;
-
-  quint32 bid = id | (data << 12);
-  // first check the cache
-  if (cache[bid] != NULL)
-    return *cache[bid];
-
-  // now find the variant
-  if (blocks.contains(bid)) {
-    QList<BlockInfo*> &list = blocks[bid];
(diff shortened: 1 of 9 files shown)


nnooney:
This pull request should help with #125

erich666:
I have nothing important to add, other than to say you're my personal hero for the week. I'm just about to start on this task for Mineways (which has diverged significantly from minutor at this point), and I suspect your explorations here will save me many hours. I'll try to post whatever new info I find here (unless you beat me to it).

nnooney:
With the update to the vanilla ids, I was able to remove most of the remaining unknown blocks for world rendering. Here is a comparison of worlds from 1.12 and 1.13:

divenon_1-12 copy

divenon_1_13 copy

These images are compressed so they will fit on GitHub, but it is interesting to see an overview of how they differ. I haven't updated the biomes file yet to add different variants for water, but that should be doable. I also noticed that the naming is wrong for `light_gray_terracotta`, so I can also fix that.



Also, here is a view of coral reefs rendered at layer 60 of this world:

world_1-13_layer60

nnooney:
I've now added the ocean biomes (and other 1.13 biomes) to vanilla_biomes.json. The oceans now have a subtle effect on the water color.

biome ocean

The colder temperatures have a darker tint applied to them to give the appearance that the water is colder. The settings could be adjusted to find the right effect; I just tinkered with them until it looked reasonable.

nnooney:
This is not backwards compatible. I had to make significant changes to the Definition Manager and Block Identifier in order to support the transition from block ids to names. I would suggest making this a separate release and continue to have the 1.12x branch available for the time being.

EtlamGit:
Your 1.12 image is created with an old version of Minutor. Biome depending Grass colors are missing, but are used for the 1.13 image.

EtlamGit:
@nnooney where did you find the values for watermodifier for the different ocean types?

nnooney:
I made up the water modifier values until I found something that I liked. They run in a gradient from blue (results in no change in color) to red (makes the water a darker color). I ordered the oceans from what I thought was warmest to coldest:

- Warm Ocean
- Deep Warm Ocean
- Lukewarm Ocean
- Deep Lukewarm Ocean
- Ocean
- Deep Ocean
- Cold Ocean
- Deep Cold Ocean
- Frozen Ocean
- Deep Frozen Ocean

These numbers can definitely be modified (I was initially hoping for a way to make the warm oceans a lighter color than the default).

nnooney:
The 1.12 image was created from the latest download available on the website, which is why it seems there is so much difference.

EtlamGit:
Aha, good to know with the water colors. The swamp watermodifier is origin directly from Minecraft source code. And as Mojang also seems to modify water color across the new ocean biomes I would like to test the color values they use in the code.
But that can wait some month...

nnooney:
I agree that using the Minecraft provided values would be the end goal here. I wonder if those values are encoded in the new biome data. I've been focusing mostly on rendering the blocks, so sometime in the future I'll take a look at where those values come from.

PR #130, opened 2018-07-30
backwards compatible 1.13 translation layer

EtlamGit:
This is the first approach to be able to load old and 1.13 version maps with the same version of Minutor.

At least here it is working with different maps. Some trillion block definitions are still missing, as Mojang renamed also a lot of blocks: https://minecraft.gamepedia.com/1.13/Flattening#Block_and_Item_IDs
I have added them for grass_block, tall_grass and vine just for test purpose. As they are highlighted and named in Minutor it should be easy to rule that out.

There are several other drawbacks in the new 1.13 definitions and code. I will try to address some of them in the rest of the year, but for now we have same starting point.
* definitions are way from complete, also colors for added blocks are wrong
* performance is very slow, this is probably because of all that strings. I will try to replace that with hashes.

That is all for now, as internet is very slow here on the open ocean...

### DIFF
--- a/chunk.cpp
+++ b/chunk.cpp
@@ -1,6 +1,22 @@
 /** Copyright (c) 2013, Sean Kasun */
 
 #include "./chunk.h"
+#include <algorithm>
+
+quint16 getBits(const unsigned char *data, int pos, int n) {
+  quint16 result = 0;
+  int arrIndex = pos/8;
+  int bitIndex = pos%8;
+  quint32 loc =
+    data[arrIndex]   << 24 |
+    data[arrIndex+1] << 16 |
+    data[arrIndex+2] << 8  |
+    data[arrIndex+3];
+
+  return ((loc >> (32-bitIndex-n)) & ((1 << n) -1));
+}
+
+
 
 Chunk::Chunk() {
   loaded = false;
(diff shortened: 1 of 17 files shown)


chunk.cpp
```diff
@@ -30,46 +32,39 @@ void Chunk::load(const NBT &nbt) {
...
-    auto rawPalette = section->at("Palette");
-    cs->paletteLength = rawPalette->length();
-    cs->palette = new BlockData[cs->paletteLength];
-    for (int j = 0; j < rawPalette->length(); j++) {
-      cs->palette[j].name = rawPalette->at(j)->at("Name")->toString();
-      if (rawPalette->at(j)->has("Properties"))
-        cs->palette[j].properties = rawPalette->at(j)->at("Properties")->getData().toMap();
-    }
-    // map BlockStates to BlockData
-    // todo: bit fidling looks very complicated -> find easier code
```
nnooney: I tinkered around with the bit fiddling to remove the need to call `std::reverse` in my latest commit to #127. 

The data in BlockStates is a bit-reversed array stored in little endian format. So for a 4 block array, where each block is six bits, looks like this:

```
BBAAAAAA CCCCBBBB DDDDDDCC
```

If each block had a value of `5 (000101)`, then this example array looks like this:

```
BBAAAAAA CCCCBBBB DDDDDDCC
00101000 10001010 10100010
```

The algorithm here will byte-swap the array, so that the array looks to be in order; from there I can grab the bits by passing in the index and the number of bits I want to grab. In the example below, to grab block C, I would call `getBits(data, 6, 6)` on the byte-swapped array, which looks like this:

```
DDDDDDCC CCCCBBBB BBAAAAAA
```

`getBits` concatenates enough bytes together (1) so that I'm guaranteed the whole data I want is in the same word. It then shifts this combined word into place (2) and masks it (3) to obtain the data I want to return (4):

```
(1) DDDDDDCCCCCCBBBBBBAAAAAA <- The concatenated word
(2) 000000000000DDDDDDCCCCCC <- Shift the word so that the data is in the lowest bit
(3) 000000000000000000111111 <- The mask produced by ((1 << n) - 1)
(4) 000000000000000000CCCCCC <- Result of (2) & (3)
```

I hope this information can help lead to a more efficient algorithm. The commit I recently made changes `getBits` to operate on the little endian array so the byte-swap isn't necessary. There is definitely room to clear up this algorithm.

EtlamGit: Mojang is using Little Endian, how strange is that?
I think it is a good idea to take a look to the code they use to decode that part. Mostly they have time to develop nice and efficient algorithms. But this is also something for later this year...

Just as comment for later:
A section can only hold 16*16*16 blocks, so at maximum 4096 different blocks can reside in a palette. This leads to maximum of 12 bits that has to be supported. 12 bits can spread at maximum over 3 bytes.
Typical values are probably 3..6 bits.

nnooney: I'm curious to find out how Mojang parses the data. I would agree that using their algorithm (if possible) is the best route going forward. These patches to Minutor makes it one of the first mapping tools to support block-level information about 1.13 worlds that I know of.

Also, if each block was unique, the palette would have 4096 entries, which requires 13 bits (a one followed by 12 zeros). This is still spread at maximum over three bytes.

definitions/vanilla_biomes.json
```diff
@@ -257,25 +260,87 @@
...
-      "name": "Mesa Plateau",
+      "name": "Badlands Plateau",
       "color": "#ca8c65",
       "temperature": 2.0,
       "humidity": 0.0
     },
+    {
+      "id": 40,
+      "name": "Small End Islands",
+      "color": "#282898"
```
erich666: I found some of the new biome temperatures (if they're to be believed) here: https://minecraft.gamepedia.com/Biome#Medium.2FLush_biomes

EtlamGit: I will wait until MCP is released for 1.13 and look into the sources...

PR #134, opened 2018-08-29
complete color overhaul

EtlamGit:
This is the complete new color set based on the Vanilla 1.13.1 textures for all known 593 Blocks.
(And also for the new Entities)

I tried to develop automatic color selection rules to be able to reproduce the selected colors. This will also be a big help for new blocks in upcoming Minecraft releases:
* Use the average color of the top side texture for each block (use ImageMagick with box filter to average)
* In case there is an alpha channel in the texture, it is a bit complicated how to average:
  + To get comparable look to the game, I blended Blocks (that typically occur multiple times above each other and get darker with each Block) towards a black background (e.g. leaves and water).
  + Other transparency is used as is (e.g. decorative stuff, glass).
* Ores would vanish in gray, therefore they remain with their old color
* Flowers derive their color only from the top part (without the green)
* Stone and Gravel are modified a bit to be distinguishable from Cobblestone (5% darker/brighter).
* Stairs and Slabs of the same type than a standard Block are modified a bit to be distinguishable (5% darker/brighter).

I would appreciate if everybody takes a look into the new colors and gives some feedback about the "new" look. It should be close to the old one, but closer to the Vanilla Minecraft than before.

### DIFF
--- a/definitions/readme.txt
+++ b/definitions/readme.txt
@@ -71,7 +71,10 @@ String representation of this Block, shown in status bar when hovering mouse ove
 
 "color": "#ffffff"
 
-Color usedd for rendering this block
+Color used for rendering this block
+ For new Blocks, please use the average color of the texture Minecraft would use for the top side.
+ ImageMagick can do that for you:
+ > convert ${filename} -filter box -resize 1x1 -depth 8 txt:-
 
 
 "alpha": float_number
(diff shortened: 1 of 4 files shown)


EtlamGit:
Just to clarify:
* Taking the average top texture color is exactly the same principle rule I took in the past years
* I just moved from using Gimp to ImageMagick, as I can do it now with scripts on the command line

So Minutor should look the same than before. Just for Blocks where Mojang changes the colors in between (e.g. wool) have now the "correct" color. The only exceptions from the basic rule are listed above and are open for discussion.

PR #142, opened 2018-10-04
Restore location on reload & Add y coordinate to status bar

### LINKED ISSUE #141: Feature request: Restore location after reload and add Y to status bar.
After reloading (F5) I want the view to return to the location instead of jumping to the spawn point.
I want the block's Y coordinate in Statusbar in addition to the X and Z.

I will have a pull requests with these changes soon.

cpforbes:
Fixes #141


EtlamGit:
Changes in mapview.cpp look ok.

Changes in minutor.cpp are not obligatory, but would also not harm.
From a coding standpoint they look also fine.

PR #145, opened 2018-10-10
fix Golems

EtlamGit:
Iron & Snow Golem got new IDs, which makes now 3 differnt variants across all Minecraft versions.


EtlamGit:
I do not like the way I have implemented the (again) changed name for Snow Golems.
In the next days I will try to find a better solution.

PR #152, opened 2019-03-04
adopt 1.13.2

hezhenwei:
Seems 1.13.2 data changed some, so i modified the code to fit it.
Not perfect, but at least most block can be shown.
some unhandled block is treated as gold block so it could be easily identified and fixed in future.


hezhenwei:
sorry the pull is not include all changes. close this pull and i will do a new one

EtlamGit:
Em, you know that we implemented Minecraft 1.13 support 6 month ago?
It is still in a development branch, but I will merge it back to master this week.

PR #153, opened 2019-03-04
adoption to 1.13.2

hezhenwei:
while i was playing 1.13.2 server, some saved data cannot be shown.

Seems 1.13.2 data changed some, so i modified the code to fit it.
Not perfect, but at least most block can be shown.
some unhandled block is treated as gold block so it could be easily identified and fixed in future.


EtlamGit:
Hi @hezhenwei !

Probably you missed our development branch for Minecraft 1.13.
So sorry, but your work seems to be obsolete, as we did full support 6 month ago (all Blocks, all Entities, backwards compatible).

Nevertheless I will take a look into your changes. Maybe some nice ideas could get adopted.

hezhenwei:
Hi EtlamGit...

Yes, I noticed that just after I create this pull request...
I wrote to mrkite's email to explain this. 
It's sad :(
Anyway it's greate to know that you have official update.
BTW, can I participate your team to update minutor?
Testing, compiling etc., any kind of work.

PR #159, opened 2019-04-23
Update 1.14

EtlamGit:
Unfortunately Mojang changed storage format again, Minutor crashes with latest occurrence of negative and empty Sections inside Chunks. So we have to do a new release.

+ many new Slabs, Stairs and Walls
+ Panda, Bamboo and Biome Bamboo Jungle (Hills)
+ many new working blocks
+ and much more

### DIFF
--- a/chunk.cpp
+++ b/chunk.cpp
@@ -81,16 +81,19 @@ void Chunk::load(const NBT &nbt) {
     int numSections = sections->length();
     // loop over all stored Sections, they are not guarantied to be ordered or consecutive
     for (int s = 0; s < numSections; s++) {
-      ChunkSection *cs = new ChunkSection();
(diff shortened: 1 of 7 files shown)


mrkite:
I suppose we should prep a new release after this..  or do you think we should wait to see what else they break?  (I'll also take that opportunity to roll back the Qt version on the Mac)

EtlamGit:
I'm confident, that I found and fixed everything that was broken this time. Maybe we could wait with a release for 1..2 more weeks.

Today I tried to fix the Qt 5.12 issue.

didaquis:
Hi @mrkite! 

Can you make a new release of Minutor? 

Thanks to you, @EtlamGit and other contributions for your amazing work!!

PR #160, opened 2019-04-24
fix hashing of definition file names for Qt 5.12.x

### LINKED ISSUE #149: builds with Qt 5.12.0 are not working
It seems that Qt 5.12.0 (or some other version between LTS 5.9.7 - 5.12.0) has changed the behavior of qHash(QString). We get multiple different results when hashing the definition file name. And therefore start to pollute the folder `%localappdata%/seancode/Minutor/`.

Solution: stick to LTS 5.9.x for the moment.

EtlamGit:
This would close #149 
Please review the changes carefully !

- Main fix is to add a specific seed (not zero) when doing the hashing during creating file names for the definitions. I took 42 as arbitrary number.
- Remaining problem was that old definition files are still stored in user data folder and registry when first starting the new Minutor version. I created a fixed list with these 5 names and force-delete them before loading anything else. This is not the smartest way, but maybe someone comes up with something better...

### DIFF
--- a/definitionmanager.cpp
+++ b/definitionmanager.cpp
@@ -74,11 +74,18 @@ DefinitionManager::DefinitionManager(QWidget *parent) :
   QSettings settings;
   sorted = settings.value("packs").toList();
 
+  // clean old hashed files without extra seed
+  QString destdir = QStandardPaths::writableLocation(QStandardPaths::DataLocation);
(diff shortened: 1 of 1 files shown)


mrkite:
Makes sense.  A long-term solution can come later.

PR #162, opened 2019-05-05
repair of definition files

### LINKED ISSUE #161: Error hashing block (each blocktype)
Windows 64, release 2.2.0 (noinstaller zip). Upon starting, the program displays hundreds of message boxes, each saying "Error hashing block" + blocktype, for each and every blocktype. After discarding each message box, I can open worlds without any apparent problems.

I had an m1.13-branch self-compiled binary working previously, now replaced with the official release executable. I tried clearing all the definitions in `%localappdata%/seancode/Minutor`, that doesn't seem to help.

EtlamGit:
This also closes #161

Added an new private method to DefinitionManager to hold all maintenance code regarding definition files. This contains for the moment:
* force delete old hashed names (before Qt5.12 compatibility)
* delete orphaned definitions in settings (matching files were deleted)
* refresh old definitions on disk with build-in definitions (from newer releases)

### DIFF
--- a/blockidentifier.cpp
+++ b/blockidentifier.cpp
@@ -261,7 +261,7 @@ void BlockIdentifier::parseDefinition(JSONObject *b, BlockInfo *parent,
     // and prevents generating a wrong definition file
     QMessageBox::warning((QWidget*)(NULL),
                          "Error loading Block definition: " + name,
-                         "Failed to add Block from definition file, as it might be a duplicate\nor generates the same hash than an already existing Block." ,
+                         "Failed to add Block from definition file, as it might be a duplicate\nor generates the same hash as an already existing Block." ,
                          QMessageBox::Cancel, QMessageBox::Cancel);
   }
   blocks.insert(hid, block);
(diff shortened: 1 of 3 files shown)


blockidentifier.cpp
```diff
@@ -264,6 +265,7 @@ void BlockIdentifier::parseDefinition(JSONObject *b, BlockInfo *parent,
                          name,
                          QMessageBox::Cancel, QMessageBox::Cancel);
   }
+#endif
```
madmaxoft: I think that rather than making this debug-only, the error message should be a bit more descriptive, while keeping it in the release version as well (so that it checks other authors' definition files as well).

How about `Failed to add block %1 definition from file %2, because it is already present in the loaded definitions.` ? I know it's not exactly what's happening, but it's still of much more help than the old message.

EtlamGit: File name is not available at that code position.

As the old text was only for me, it is obviously not descriptive. I tried a rephrase...

blockidentifier.cpp
```diff
@@ -260,8 +260,8 @@ void BlockIdentifier::parseDefinition(JSONObject *b, BlockInfo *parent,
     // this will only trigger during development of vanilla_blocks.json
     // and prevents generating a wrong definition file
     QMessageBox::warning((QWidget*)(NULL),
-                         "Error hashing Block",
-                         name,
+                         "Error loading Block definition: " + name,
+                         "Failed to add Block from definition file, as it might be a duplicate\nor generates the same hash than an already existing Block." ,
```
madmaxoft: `than` -> `as` and I'm happy :)

Issue #169, opened 2019-08-05
MacOS version crashes on launch; Qt build errors


Down10:
I kept having the released version of Minutor crash on launch or if it does load, it crashes upon loading a game file. Example crash log:

```
Thread 6 Crashed:: Thread (pooled)
0   libsystem_platform.dylib      	0x00007fff72e62dd0 _platform_memmove$VARIANT$Haswell + 240
1   com.seancode.minutor          	0x00000001025682df Chunk::loadSection1519(ChunkSection*, Tag const*) + 2239
2   com.seancode.minutor          	0x0000000102567191 Chunk::load(NBT const&) + 1713
3   com.seancode.minutor          	0x000000010256a7a1 ChunkLoader::run() + 801
4   com.seancode.minutor          	0x0000000102dd545d QThreadPoolThread::run() + 125
5   com.seancode.minutor          	0x0000000102dd1433 QThreadPrivate::start(void*) + 323
6   libsystem_pthread.dylib       	0x00007fff72e6e2eb _pthread_body + 126
7   libsystem_pthread.dylib       	0x00007fff72e71249 _pthread_start + 66
8   libsystem_pthread.dylib       	0x00007fff72e6d40d thread_start + 13
```

I am not a C++ developer, but I thought I'd try building the app from the original source myself on my own iMac [Retina 5k 2019; Mojave 10.14.5; 3.1 GHz Intel Core i5; 16 GB RAM] to see if that could work. I installed the open source Qt Creator with WebEngine [v4.9.2; Based on Qt 5.12.4 (Clang 10.0 (Apple), 64 bit)] and opened up the latest Minutor code from GitHub, which then displayed a bunch of code warnings and at least one error:

<img width="1372" alt="minutor pro errors" src="https://user-images.githubusercontent.com/91656/62493576-54ef9980-b786-11e9-8805-6ecdc74c54d9.png">

Perhaps these issues can be addressed to get the Mac app running smoothly again. And if there is a smarter way to build from the source, let me know. Thanks.

Down10:
Compile output log, failed with errors: https://pastebin.com/fFtwRJbq

madmaxoft:
Thanks for the compile error log, I've fixed those in #170. As for the runtime crashes, I can't help there, don't have a Mac.

kdbertel:
I also was getting crashes trying to run the latest release (2.2.0) on macOS (10.14.6). Is there any chance these might have been fixed in the various fixes and improvements since that release?

[...]

PR #170, opened 2019-08-06
Fixed clang compile error and some warnings.

madmaxoft:
Fixes the compiler error mentioned in #169


Condor987:
I see that errors that make a mac crash have supposedly been fixed... New to GitHub here and I'm wondering how to download the fixed version?

Issue #171, opened 2019-08-06
Unify indent, add editorconfig


madmaxoft:
The project uses several indent styles, it'd be great to pick one and use an [editorconfig](https://editorconfig.org) file to enforce it in compatible IDEs.

madmaxoft:
I could add the file, but first we need the decision on the indent style - tabs or spaces (how many).

EtlamGit:
As far as I remember, we used tabs only far in the past.
After switching to Google linted code we now have 2 spaces.
So there are probably mixed files left. I can not remember seeing anything like 3 or 4 spaces.

I think this editorconfig file will not help directly:
- This project is set up with Qt Creator and that editor is not on the list of editors that support this type of configuration.
- Class declarations uses 1 space indent for public/private. This even now does not match the 2 space rule.

But as it does not harm to have such a file, it may help in the future.

mrkite:
Yeah, it used to be just chaos so I decided to reformat everything using cpplint with google's default ruleset.

I'm not too picky about things google is picky about, but 2 spaces is definitely the way to go for indenting.. and the public/private/slots/signals tags should either be 1 space indented or not indented at all.

One other thing we should probably do is switch to the qt5 signal/slot format.

https://wiki.qt.io/New_Signal_Slot_Syntax

This allows compile-time checking of the signals and slots, and you don't have to provide all the argument types anymore.  The only drawback is you cannot connect directly to a slot that takes an optional argument without providing a signal that passes that argument.  (You can easily use a lambda to get around this though).

madmaxoft:
There's a plugin for QtCreator to support editorconfig, it's not yet listed on the homepage, but it works: https://github.com/editorconfig/editorconfig-qtcreator

This would help me tremendously, I use tabs for my other projects, and having to convert tabs back to spaces before committing to Minutor is a chore.

I've seen 4-space indents in Minutor, off the top of my head in `DefinitionManager`'s constructor initialization list. The 1-space indent of class access specifiers is weird :)

cpforbes:
Here is the QT coding style: https://wiki.qt.io/Qt_Coding_Style for reference.

PR #172, opened 2019-08-06
Added basic editorconfig file.

madmaxoft:


madmaxoft:
As discussed in #171, this only adds the editorconfig file, but doesn't touch the indents.

PR #173, opened 2019-08-06
use shared pointer inside Chunk Cache

EtlamGit:
In very seldom cases Minnutor still crashes with memory access
violations:
If a Chunk gets removed from the Cache but still gets loaded or
rendered, the data structure is removed from heap, but still written to.

By using QSharedPointer<Chunk> always some routine has ownership and the
heap is only freed when all routines are finished with processing.

### DIFF
--- a/chunkloader.cpp
+++ b/chunkloader.cpp
@@ -4,11 +4,13 @@
 #include "./chunkcache.h"
 #include "./chunk.h"
 
+
 ChunkLoader::ChunkLoader(QString path, int cx, int cz)
   : path(path)
   , cx(cx), cz(cz)
   , cache(ChunkCache::Instance())
 {}
+
 ChunkLoader::~ChunkLoader()
 {}
 
@@ -44,7 +46,7 @@ void ChunkLoader::run() {
     return;
   }
   // get existing Chunk entry from Cache
-  Chunk *chunk = cache.fetchCached(cx, cz);
+  QSharedPointer<Chunk> chunk(cache.fetchCached(cx, cz));
   // parse Chunk data
   // Chunk will be flagged "loaded" in a thread save way
   if (chunk) {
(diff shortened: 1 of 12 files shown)


chunkloader.cpp
```diff
@@ -44,7 +46,7 @@ void ChunkLoader::run() {
     return;
   }
   // get existing Chunk entry from Cache
-  Chunk *chunk = cache.fetchCached(cx, cz);
+  QSharedPointer<Chunk> chunk(cache.fetchCached(cx, cz));
```
madmaxoft: How about using `auto` more?
`auto chunk = cache.fetchCached(cx, cz));`
Doesn't matter then if it's a `QSharedPointer`, `std::shared_ptr`, naked pointer or whatever else.

madmaxoft:
With this kind of change, it is critical to check that the memory is actually freed eventually. It's easy for the shared pointers to keep things alive forever.

chunkloader.cpp
EtlamGit: I don't like auto that much, as you do not see the type of a variable in a normal editor.
(I know that this answer is contradictory to your (good) argument. Probably it is just personal style how much auto you use.)

EtlamGit:
I observe normal RAM usage. But you are right, we should test this very well.

With the old code I still got (seldom) crashes. So there is need to change something. With the shared pointers I'm not able to reproduce these crashes at the moment.

mrkite:
Just need to make sure we're not holding onto any qsharedpointers for no reason.

EtlamGit:
There are 3 locations where these shared pointers are used:
- In normal code to insert it into the Cache or work on Chunk data from the Cache. In these cases, the local shared pointer variable is deleted when leaving the local scope of the function. If something goes wrong here, the complete application would get stuck.
- In worker threads to load or render Chunks. These are deleted when the task is finished. If this would fail, our worker pools would get stuck.
- And obviously inside the Cache itself. Qt claims to delete objects that get thrown out of the Cache. And I did observe that behavior.

So in my opinion we would notice quite fast if something is going wrong.

PR #180, opened 2019-08-22
Fixed crash on unknown biome.

### LINKED ISSUE #179: Weird crash on Win, 1.14 world
I'm getting a really weird crash. I compiled commit c046fe8a8a1cce1979850931dfd21d299afe70f7 against Qt 5.11 static 64-bit MSVC build (my usual Minutor setup) and I get a crash, trying to access a nullptr location in the renderer:
```
minutor.exe!QHash<int,QList<BiomeInfo *> >::operator[](const int & akey) Line 756
minutor.exe!BiomeIdentifier::getBiome(int biome) Line 212
minutor.exe!ChunkRenderer::renderChunk(QSharedPointer<Chunk> chunk) Line 41
minutor.exe!ChunkRenderer::run() Line 25
minutor.exe!QThreadPoolThread::run() Line 101
```
The weird thing is it crashes inside Qt code, accessing a `QHash`-ed value, as if the `BiomeIdentifier::Instance().biomes` object wasn't valid. Even weirder, it only happens when built in release mode (with or without debugging info doesn't matter), it doesn't happen in debug builds. And it only happens for a specific place in my map:
http://xoft.cz/minutor/1_14.7z (25 MiB), chunk [-26, 37].
The world normally renders fine in the default window size, only when maximized to fullscreen a moving around a bit the crash happens.

madmaxoft:
Fixes #179.

### DIFF
--- a/biomeidentifier.cpp
+++ b/biomeidentifier.cpp
@@ -118,7 +118,7 @@ QColor BiomeInfo::mixColor( QColor colorizer, QColor blockcolor )
 }
 
 
-QColor BiomeInfo::getBiomeGrassColor( QColor blockcolor, int elevation )
+QColor BiomeInfo::getBiomeGrassColor( QColor blockcolor, int elevation ) const
 {
   QColor colorizer;
   // remove variants from ID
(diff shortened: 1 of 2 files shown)


madmaxoft:
This is a hotfix for the problem so that it doesn't seem to happen anymore, but still technically the code is broken, because `QHash` is not guaranteed to be thread-safe. On the other hand, locking the hash for querying biome in every pixel would be too much of a performance hit.

EtlamGit:
Could you please also make this method const?

madmaxoft:
Here you go.

EtlamGit:
Great! And you found even more methods to be set to const.

madmaxoft:
`const` usually grows like cancer, they had to be const-ed in order to use the const return value.

PR #186, opened 2019-09-15
support zooming "out" - below zoom factor 1.0.

cre4ture:
Works using QPainter::drawImage functionality instead of manual memcpy for displaying pre-rendered chunk images.


EtlamGit:
The most important zoom factor for Minutor is 1:1 (one Minecraft Block per screen pixel).
If you zoom IN, you get more pixels per block but still see all details (fine).
If you zoom OUT, you will loose information, as multiple blocks get blended (bad).


First comment:
How do you ensure that it is easy for the user to get exactly this 1:1 zoom?
This is an UI question. What do we implement to achieve good usability of this feature?
At the moment 1:1 is at one end of scrolling direction. This is obviously easy to use. Just scroll into one direction. It does not matter if you scroll too much.


Second comment:
How do we ensure the users PC has enough RAM for more Chunks on the screen?
If you zoom OUT too much without enough RAM, Minutor will stall or crash.


Don't get me wrong, I also would like to zoom out more than possible at the moment. But during the last years I did not find answers to the questions above...

cre4ture:
OK, I got your points.
1 - Yes, exact 1:1 zoom can't be achieved anymore after zooming afterwards. That's something I would suggest to fix by introducing a shortcut key combination that resets to zoom factor 1.0. For example Ctrl+0 like many users are already used from other applications like browser.
2 - Yes, availability of RAM can be an issue, depending on users pc hardware.
I saw in the code that there is already some logic implemented to check how much free RAM is available. It's part of the ChunkCache. I would suggest to use this information, combined with window size to determine automatically the possible zoom factor range.

Apart from that, do you think it would be OK to just introduce an "experimental" menu entry which allows to enable such features like this? I think that many users probably still can profit from it - even though it's not yet perfect in all details.

EtlamGit:
1) Never change working behavior that is expected by the users.
So I would like to stick to the known behavior/procedure to get to exactly 1:1 zoom factor (scrolling/zooming out until it stops). What is about using a modifier key for the zoom-out-more feature?

2) The RAM estimation code is not working perfectly at the moment. But it should be improved to define the maximum possible zoom-out value. Maybe number of Chunks needed for maximum-zoom-out at full-screen plus some margin for scrolling (e.g. 25%).

3) There is already a dialog for configuration: Help->Settings
We could place some mode/enable stuff there.

EtlamGit:
Is it possible that your mouse wheel creates much more scroll events than typical ones?

I just compared your "more-natural" zoom behavior to the current one:

- at the moment we need 16 scroll events (15° turns) to get to 20x zoom (maximum)
- your solution needs 262 scroll events, which takes eternities for me to reach

Clipboard01


By calculating the real zoom factors in Excel, I come to the conclusion, that simple adding a one when the mouse has turned 15° (120 delta) would be much better. In that case we always have integer multiples of a single block, which would generate the least distortions during rendering.

cre4ture:
The new feature forced me to introduce a change regarding the zooming. Consider moving below zoom factor of 1.0, then the approach of just adding/substracting 1 simply doesn't work anymore. 1-1 = 0

But yes, I guess the new approach is much "slower" when compared with the original implementation. The reason why I think the suggested approach is more "natural" is that the relative step size is always the same. Humans perceive the reality in logarithmic scales and thus I thought it's more natural.
But anyway I think I can give also the new zoom a bit more speed and then update the pull request.

One question about the "least distortions during rendering" you mentioned: Is your concern in this regards the performance during drawing or the visual appearance?
If its performance then this might be a stop for this pull request, as I just rely on the QPainter's implementation and if this isn't acceptable for you, I will not be able to improve there anything.

mrkite:
@EtlamGit I've found a trackpad with smooth scrolling and pinch-to-zoom sends out far more events than a mousewheel... that could be what they're seeing?

EtlamGit:
About distortion:
There are many map creators and creative builders and out there, which are planning their builds with the help of a map. For these people, it is essential that a Block in Minecraft is displayed on the map always with the same scale. Even distances and relations have to be "mapped" correctly.
If we have non-integer zoom factors, this is not ensured, as e.g. Block 1 and 2 are mapped to pixel 1 but Blocks 3 to 5 to pixel 2. You can find more elaborated description of this effect when you search for "aliasing".

Even the current implementation is not ensuring this. Therefore, I suggested to find a new zooming behavior that achieves this. You are right, something logarithmic is more natural, Fibonacci would also work, or squared like you did. But I think at the end this curve should be mapped to integer numbers.

As example for zoom-in: 
 - always add +1 for each 15° scroll event to zoom_internal
 - use a squared or other nice mapping to get real zoom
 - round to next integer

Example for zoom-out:
 - always add +1 to another internal variable
 - use 1/variable to create zoom factors 1/2 1/4 1/8 1/16
 - it is impossible to zoom-out more than 16, because than a complete Chunk will be mapped to one pixel

As a real-live example from my map: I have a build structure that has a border of 1 Blocks. These Blocks have a high contrast to the surrounding (Granite in Water). When you now start zooming-out this look very ugly, as sometimes these lines are partly vanishing and coming back during zooming. You can test that also with a large white square on black background. With exact integer zoom factors (1/integer) these effects look much less worse.


About performance:
I did not test performance impact. But I believe the QPainters implementation might be even faster than the current one. And if not, it is not rocked-science to write own blending/rendering code.

cre4ture:
Thanks for your patience and explanations, I didn't know all of the main usecases this tool is used for.
I updated the pull request with a change that reintroduces the previous zoom behavior as the default zoom behavior. I added two entries in the setting dialog.

One which allows to enable the "more natural" zoom approach called fine zoom now. Actually I "speed" up this zoom behavior quite a lot. Thanks for your input (testing and comments) there. The other option is to enable the "zoom out" feature which allows a minimal zoom factor of 0.20 (this was the useful limit on my hardware). The description of this option contains also the hint that it would need quite some RAM. Both settings are in a new QGroupBox called "experimental".

I like your additional ideas regarding further zoom improvements. Anyway I suggest that this should be done then in other pull requests, as it is currently not of a priority for me as I'm not doing creative building or map creating. Rather I use the map for faster exploration.

EtlamGit:
I will test your changes as soon as possible, but unfortunately I'm short of time at the moment...

You idea to split topics to different Pull Request is fine. I planned to add something to restrict zoom levels anyway to fix an issue regarding out of memory and resulting crash.

EtlamGit:
Finally I found some time to do first tests:

- The new zoom behavior is nice. I think we can use a non linear zoom as default. I will do that when developping the changes regarding memory issues caused by zoom-out.

- We still have rendering issues when zooming-out fine structures:
Your code: grafik with bilinear filtering: grafik
As shown in the second image, it gets better when doing some blending/filtering (combining all relevant pixels from the Chunk to a screen pixel).

This can be achieved by changing mapview.cpp#428:
```
  QPainter canvas(&imageChunks);
  if (this->zoom < 1.0)
    canvas.setRenderHint(QPainter::SmoothPixmapTransform);
  canvas.drawImage(targetRect, srcImage);
```
You might want to add that change to your Pull Request. Otherwise I would do that during Christmas.

- Regarding Performance, I was not able to notice any difference to the old code. But my assumption is that QPainter:drawImage() is using most efficient code and probably even GPU support for doing the task. So it should be faster than our old pixel-per-pixel copy.

- As all changes from this Pull Request are disabled by default, I think it is save to merge them.

cre4ture:
@EtlamGit Nice improvement. I like it too. I put the code snippet into the pull request as suggested.

EtlamGit:
@mrkite I did not test deeply, but it seems to work.
As all changes from this Pull Request are disabled by default, I think it is save to merge them.

I will use the new zoom behavior as basis to fix out-of-memory issues.

Issue #192, opened 2019-11-27
Minutor v2.3.0 crashes while scrolling around map


archiesholland:
Minutor v2.3.0 crashes while scrolling around map.

[Report.wer.txt](https://github.com/mrkite/minutor/files/3894765/Report.wer.txt)
[Report1.wer.txt](https://github.com/mrkite/minutor/files/3894766/Report1.wer.txt)

EtlamGit:
Are you using the latest build or an ancient release?

archiesholland:
Latest build, from https://www.dropbox.com/s/c9n320hjet3hdjf/minutor.zip?dl=0

Archie S. Holland
Archie/Brother/Dad/Uncle/Grandpa
archiesholland@gmail.com
home: 270-765-2530
cell: 270-763-7530

"Shared pain is diminished - shared joy is increased"
------Mike Callahan

EtlamGit:
The nightly changed 2-3 days ago. Do you use that new version?

Is this crash only for one special world? Or even for new created worlds?
If only on world, please provide the world,

On what zoom level do you scroll and crash?
Do you use the new zoom out feature. (still unstable, so disable for the moment)

What about your RAM and overall memory usage?
Do you use a 64bit operation system?
32bit Windows will not provide enough memory for larger window size.
Is your monitor too large or your RAM too small?

archiesholland:
Yes, I'm using that newest version.
I've only used it on one world so far. (attached)
Using the default zoom level (didn't know there WAS a new zoom out feature)
(grin)
Windows 10 Pro
Version 10.0.17763 Build 17763
Dell Inc. Inspiron 5558
x64-based PC
Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz, 2401 Mhz, 2 Core(s), 4 Logical
Processor(s)
Installed Physical Memory (RAM) 8.00 GB
Total Physical Memory 7.91 GB
Available Physical Memory 4.13 GB
Total Virtual Memory 10.3 GB
Available Virtual Memory 5.46 GB
Page File Space 2.38 GB
NVIDIA GeForce 920M
Intel HD Graphics 5500

Just noticed that it only seems to occur at edges of the map.

 World36.zip
<https://drive.google.com/file/d/10ql51hQY4lq9FWDvHiuTg7xVDRCDRhwZ/view?usp=drive_web>

Archie S. Holland
Archie/Brother/Dad/Uncle/Grandpa
archiesholland@gmail.com
[...]

PR #195, opened 2019-12-10
fix potential stability problem

cre4ture:
shared pointer to cached chunk needs to be _copied_ from cache before unlocking the mutex

### DIFF
--- a/chunkcache.cpp
+++ b/chunkcache.cpp
@@ -55,9 +55,9 @@ ChunkCache& ChunkCache::Instance() {
 
 void ChunkCache::clear() {
   QThreadPool::globalInstance()->waitForDone();
-  mutex.lock();
+
+  QMutexLocker guard(&mutex);
   cache.clear();
-  mutex.unlock();
 }
 
 void ChunkCache::setPath(QString path) {
@@ -80,9 +80,9 @@ int ChunkCache::getMaxCost() const {
 QSharedPointer<Chunk> ChunkCache::fetchCached(int cx, int cz) {
   // try to get Chunk from Cache
   ChunkID id(cx, cz);
-  mutex.lock();
+
+  QMutexLocker guard(&mutex);
   QSharedPointer<Chunk> * p_chunk(cache[id]);   // const operation
-  mutex.unlock();
 
   if (p_chunk != NULL )
     return QSharedPointer<Chunk>(*p_chunk);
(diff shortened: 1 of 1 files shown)


cre4ture:
might fix #192

chunkcache.cpp
```diff
@@ -80,9 +80,9 @@ int ChunkCache::getMaxCost() const {
...
   // try to get Chunk from Cache
   ChunkID id(cx, cz);
-  mutex.lock();
+
+  QMutexLocker guard(&mutex);
   QSharedPointer<Chunk> * p_chunk(cache[id]);   // const operation
-  mutex.unlock();
 
   if (p_chunk != NULL )
     return QSharedPointer<Chunk>(*p_chunk);
```
cre4ture: this copy operation here needs to be done within the mutex protected scope. otherwise the cache could have removed the shared pointer instance which p_chunk is pointing to already before.

chunkcache.cpp
```diff
@@ -93,22 +93,26 @@ QSharedPointer<Chunk> ChunkCache::fetchCached(int cx, int cz) {
...
-  if (p_chunk != NULL ) {
-    QSharedPointer<Chunk> chunk(*p_chunk);
-    if (chunk->loaded)
-      return chunk;
-    return QSharedPointer<Chunk>(NULL);  // we're loading this chunk, or it's blank.
+  {
+    QMutexLocker guard(&mutex);
+    QSharedPointer<Chunk> * p_chunk(cache[id]);   // const operation
+    if (p_chunk != NULL ) {
+      QSharedPointer<Chunk> chunk(*p_chunk);
```
cre4ture: same here

chunkcache.cpp
```diff
@@ -127,5 +131,7 @@ void ChunkCache::routeStructure(QSharedPointer<GeneratedStructure> structure) {
 void ChunkCache::adaptCacheToWindow(int wx, int wy) {
   int chunks = ((wx + 15) >> 4) * ((wy + 15) >> 4);  // number of chunks visible
   chunks *= 1.10;  // add 10%
+
+  QMutexLocker guard(&mutex);
```
cre4ture: setMaxCosts might also (if number decreases) erase elements from cache and thus needs to be protected

chunkcache.cpp
```diff
@@ -55,9 +55,9 @@ ChunkCache& ChunkCache::Instance() {
 
 void ChunkCache::clear() {
   QThreadPool::globalInstance()->waitForDone();
-  mutex.lock();
+
+  QMutexLocker guard(&mutex);
```
cre4ture: using QMutexLocker is in general the more robust apporach as it also unlocks the mutex in case of an exception or a return.

EtlamGit: That is obviously right.

EtlamGit:
Binaries are updated, I will test tomorrow.
Thanks for the help.

PR #196, opened 2019-12-14
vanilla_ids: Fixed JSON lint errors, formatted.

madmaxoft:
Ran the vanilla_ids.json through a JSON linter, it complained about two extra commas. Fixed those, and used the linter as a formatter, too (thus the `0.0` -> `0` conversion)

### DIFF
--- a/definitions/vanilla_ids.json
+++ b/definitions/vanilla_ids.json
@@ -7,7 +7,7 @@
       "id": 0,
       "name": "Air",
       "color": "#ffffff",
-      "alpha": 0.0,
+      "alpha": 0,
       "transparent": true,
       "spawninside": true
     },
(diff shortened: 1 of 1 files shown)


EtlamGit:
The linter is right with the commas and spaces.
As alpha is a floating point number, 0 and 0.0 should both be fine. There are even languages that throw an error if you give 0 as floating point number. So, I would disagree with the linter configuration setting in that case. But as it does not really matter in our case I'm also fine with both results.

I typically use [jsoneditoronline.org](https://jsoneditoronline.org/) to verify larger changes.
What tool are you using?

madmaxoft:
The one built into DDG: https://duckduckgo.com/?q=json+lint

PR #200, opened 2019-12-18
Fix crash in worldsave when loading incompletely written chunks

EtlamGit:
This is the same bug than in #197 but in WorldSave::run() during writing a PNG.

As we have duplicated code here, which is hard to maintain. Therefore I refactored and tried to use only one place to load the NBT data for a Chunk.

Maybe @cre4ture can review these changes also.


cre4ture:
@EtlamGit: for me this looks good

PR #201, opened 2019-12-25
Restrict zoom based on available memory

EtlamGit:
These changes restrict ZOOM based on available physical RAM and current window size. It should fix all out of memory issues we have at the moment.
Basically possible ChunkCache size is estimated at startup. Only a conservative amount of Chunks is reserved at that moment. Each time the user changes the window size or zooms, the ChunkCache is increased until we reach the initially estimated maximum. In case we hit that boundary, ZOOM is decreased automatically until the amount of visible Chunks matches the Cache again.

To achieve this, the ZOOM behavior was changed to:
* a linear INDEX that is changed by the Mouse Wheel in increments of 1 for each typical scroll event
* a zoom curve TABLE to map a desired zoom factor from the given index
  * Positive indices result in the well known zoom IN but with something like the new proposed exponential behavior. As it is implemented via a table-look-up any desired behavior is possible. I took Fibonacci numbers to achieve only integer multiples of Blocks. 
  * Negative indices result in the new zoom OUT feature. Values are mapped like before:
    -1 -> 2^1 -> 2:1
    -2 -> 2^2 -> 4:1
    -3 -> 2^3 -> 8:1
    -4 -> 2^4 -> 16:1 one Chunk per pixel (maximum possible) demanding about 32..64G free RAM

The settings for experimental zoom out are removed, as it is working like desired.
Instead I added new settings to change the modifier keys which are used to determine what to do when the Mouse Wheel is scrolled. Default behavior is:
* NO modifier key: normal zoom without the possibility the accidentally zoom out
* CTRL modifier: zoom and also allow zoom out
* SHIFT modifier: change the Depth Slider

Via keyboard, zoom out is always possible.

Please carefully review the changed zoom behavior.
Restricted zoom is something we can not circumvent anyway.

### DIFF
--- a/chunkcache.cpp
+++ b/chunkcache.cpp
@@ -3,6 +3,7 @@
 #include "./chunkcache.h"
 #include "./chunkloader.h"
 
+
 #if defined(__unix__) || defined(__unix) || defined(unix)
 #include <unistd.h>
 #elif defined(_WIN32) || defined(WIN32)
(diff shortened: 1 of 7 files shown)


mrkite:
We might want to look at QEvent::TouchBegin etc for pinch to zoom etc in addition to the other zoom behaviour.

EtlamGit:
I do not have any touch devices running a Minutor build.

Another feature would be to zoom centered at current mouse position. This is the default behavior for Google Maps and other mapping applications users may be familiar with. Currently we zoom in the center of the window.

PR #202, opened 2020-01-03
Travis / Appveyor CI

### LINKED ISSUE #182: Add Travis / Appveyor CI
The CI should be used to detect PRs that don't even compile. Additionally, they can be used to provide nightbuilds.

Kolcha:
added AppVeyor and Travis CI config files

[AppVeyor build configurations](https://ci.appveyor.com/project/Kolcha/minutor):
- Visual Studio 2017, Qt 5.13.2, 64-bit
- MinGW 7.3.0, Qt 5.13.2, 64-bit

[Travis CI build configurations](https://travis-ci.org/Kolcha/minutor):
- Ubuntu 18.04, GCC 7, Qt 5.9.5
- macOS 10.14, Xcode 11.3, Qt 5.13.2

Closes #182

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,31 @@
+image: Visual Studio 2017
+
+environment:
+  matrix:
+    - compiler: msvc
+      qt: 5.13.2\msvc2017_64
+
+    - compiler: mingw
+      qt: 5.13.2\mingw73_64
+
+install:
+  # download jom for MSVC build
+  - if %compiler% == msvc (
+      curl -O -J -L -s -S http://download.qt.io/official_releases/jom/jom.zip
+      & 7z x *.zip -oC:\jom > nul
+      & del *.zip
+    )
+
+before_build:
+  - set src_path=%cd%
+  - call "C:\Qt\%qt%\bin\qtenv2.bat"
+  - if %compiler% == msvc call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvars64.bat"
+  - if %compiler% == msvc set PATH=C:\jom;%PATH%
+  - cd %src_path%
+
+build_script:
+  - qmake -config release minutor.pro
+  - if %compiler% == msvc jom
+  - if %compiler% == mingw mingw32-make -j4
+
+test: off
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+language: cpp
+
+os:
+  - linux
+  - osx
+
+dist: bionic
+osx_image: xcode11.3
+
+addons:
+  apt:
+    packages:
+      - qtbase5-dev
+      - qttools5-dev-tools
+
+  homebrew:
+    packages:
+      - qt
+
+before_script:
+  - if [[ "$TRAVIS_OS_NAME" = "osx" ]]; then brew link --force qt; fi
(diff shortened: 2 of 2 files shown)


EtlamGit:
I think it could be beneficial to stay on a LTS branch of Qt. This would be 5.12.x for the moment.

Just some time ago we had massive problems with incompatible changes in newer Qt versions (see #160). Some other incompatibilities exist for automatic definition update and Qt version.

Kolcha:
I agreed that it is better to stick to LTS, but in case of public build servers it is not always possible at least in easy way...
for example, Qt 5.12.x is included only in "Visual Studio 2019" AppVeyor image ([details](https://www.appveyor.com/docs/windows-images-software/#qt)), but Qt itself is build for MSVC2017. in general, this is not good idea to mix code compiled using different compiler versions, especially for Windows and especially for Qt. it can even don't link. or in worst case lead to unexpected and strange results on runtime.
moreover, only few Linux distributions have Qt 5.12.x (or above), and all of them have short lifetime (like Ubuntu 19.10) or updating very frequently and don't stick on specific version for a long time (like Arch). even Debian 10 doesn't have Qt 5.12.x, it uses [Qt 5.11.x](https://packages.debian.org/buster/qtbase5-dev). only starting from Debian testing Qt 5.12.x is available in repos.
for macOS situation is even worst: there is only one easy way to get some library as binary - use Homebrew. this is some kind of try to invent something like Linux repos replacement for macOS. binaries from there can be used for building/testing purposes, but they must be used very carefully when distributing app for end users, just because the way how Homebrew works and libraries was build a lot of unexpected dependencies may appear or software "may think" that it works on some Linux system, not macOS. moreover, according to [Homebrew' database](https://formulae.brew.sh/formula/qt#default), Qt 5.14.0 must be installed, why Qt 5.13.2 was used on build server is unknown to me... looks like uncontrolled environment, again. even more, it is a big question "is it possible to install previous library version using Homebrew?". at glance Homebrew provides only latest versions, and when its DB was updated, it is impossible to install previous version.

What I made supposed to be used just to ensure that everything can be built on any of supported OS. not including any runtime testing, just build, nothing more.

if you want something which will provide end user distributions, build custom Qt binaries for any required OS, upload it somewhere and use these binaries on public build servers instead of preinstalled ones. I can help with Qt build process for any OS (Win/macOS/Linux), in any configuration you want.

Kolcha:
P.S> I saw `debian` folder in repo, so I assume you build deb packages. due to different Qt versions in various distros, I think code must be ready for it. I didn't analyze affected code very well, but it looks like you need to create some file with unique name. if this file must exist permanently, and before open its name must be calculated based on some known info, it is better to have own "conversion" (known info -> filename) function, which implementation is fully controlled by you.
using library generic functions like `qHash()` for such purposes is not a good idea, because they intended to be used only on runtime, not for permanent data. from version to verson algorithm of such functions may be changed, and data produced by previous function will not be equal to data produced by current. that is happened in your case as I understood.

EtlamGit:
You are right with your observations for different Qt versions across  the different operation systems. That means we have to sort out much details before automatic builds are working flawless.

For the releases we normally did static builds to circumvent use of library versions provided by the OS. 

Minutor should work with all Qt versions from the last couple of years. Only 5.12 was a bigger problem. And you are right, we should replace `qHash()` with the current Qt implementation copied into put own code to prevent this error to happen again.

Kolcha:
let's split this pull request into 2 independent:
- first one adds CI config files (this one)
- another fixes Linux build - https://github.com/mrkite/minutor/pull/206

PR #203, opened 2020-01-04
significantly improve rendering performance

Kolcha:
Significantly improve rendering performance while changing 'depth' (height).
watch these 2 short videos to see the difference: [untouched master version](https://www.dropbox.com/s/4fa05hawrhxo96o/Screen%20Recording%202020-01-04%20at%202.28.15%20PM.mov?dl=0), [my branch](https://www.dropbox.com/s/prryml350w9b5if/Screen%20Recording%202020-01-04%20at%202.29.54%20PM.mov?dl=0)

I loaded pretty small world, on big worlds difference much more significant.

### DIFF
--- a/chunk.cpp
+++ b/chunk.cpp
@@ -22,10 +22,10 @@ quint16 getBits(const unsigned char *data, int pos, int n) {
 
 
 Chunk::Chunk()
-  : loaded(false)
-  , rendering(false)
-  , version(0)
+  : version(0)
   , highest(0)
+  , loaded(false)
+  , rendering(false)
 {}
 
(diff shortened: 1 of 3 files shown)


EtlamGit:
Very interesting! I would not have guessed that we schedule Chunks multiple times...

But after thinking deeper about this, it is obvious. Each mouse-move-event, we do a complete rendering of main window and therefore also schedule all visible Chunks for rendering. If you move the mouse slowly, we get tons of redraws and may get many thousand Chunks rendered very very often (much more that twice).

So this fix will not only improve performance, it may also increase overall stability!

Issue #199, opened 2019-12-17
search for blocks with certain id (like chests or moster spawner) and entities matching certain criterions like selling an enchanted book


cre4ture:
On my featurebranch I created a search functionality for entities and blocks. For me the feature is quite usefull to speed up search for certain blocks and entities. The results are listed in a listbox and can be displayed as an overlay item on the map. You also can sort the list according to distance.
I would like to know if you @mrkite, @EtlamGit also would be interested to have that feature on the master. I was checking if someone else also already had the need for such a search. I think #146 might be also solved by this feature, as you can just search for all the moster spawner and or chests.
Other examples would be:
- Search for diamond/gold/iron_ore
- Search for villagers that buy or sell certain items
But it will be more effort to split the feature from the other things I did there and get it onto the master.
It also brings some changes and potentially increased complexity for the threadpool and the chunkcache with it. Thats why I'm asking.

https://github.com/cre4ture/minutor-1/tree/search_enchanted_book_seller
![image](https://user-images.githubusercontent.com/252806/71001522-bf71f800-20dd-11ea-88bb-4986d59fe7fb.png)

mrkite:
What happens if you search for something common, like stone?  Does it explode or handle it fine?

EtlamGit:
I personally, do not need such a feature. But that is no reason, as there are many other features in Minutor that I do not "need" (or believed not to need at a first glance).
It is always bad to fork projects when not enough maintainers are available to support both branches. Therefore it would be very beneficial, if you migrate all your ideas into Minutor master. At least as long term goal.

The biggest drawback I see is when we have very specialized search dialog and then Mojang changes something. So the search should be as generic and maintenance free as possible.

As your changes will bring some intermediate instability, I suggest this approach:

* during Chistmas I try to fix the out of memory issues
* we do an official Release (with support for 1.14 and 1.15)
* after that we hopefully have some longer time where that Release works with latest vanilla Minecraft
* during 2020 we try to integrate your search feature


Issue #20 would also be solved.

BigAlanM:
Frankly, I would love to see this.  Specifically for chests and spawners as I use Minutor to  find dungeons and portal rooms.

cre4ture:
> What happens if you search for something common, like stone? Does it explode or handle it fine?

Your right, that's something that easily can go wrong. Currently I just put a limitation for maximum number of search results to handle this somehow. But I didn't do much testing around. So there might also be more tuning needed in this regards.
[...]

PR #214, opened 2020-03-25
preparations for search feature

cre4ture:
refactoring only: extract ChunkID from ChunkCache header
I (slowly) started with bringing the search functionality to master (#199). But this needs some preparation steps. As this work will go over a timeframe of potentially months I want to push any working version in between to master as soon as possible to avoid merging conflicts later. Hope you are not annoyed by many smaller pull requests that doesn't bring functionality nor fixes.

### DIFF
--- a/chunkid.h
+++ b/chunkid.h
@@ -1,0 +1,26 @@
+#ifndef CHUNKID_H
+#define CHUNKID_H
+
+// ChunkID is the key used to identify entries in the Cache
+// Chunks are identified by their coordinates (CX,CZ) but a single key is needed to access a map like structure
+class ChunkID {
+ public:
+  ChunkID(int cx, int cz);
+  bool operator==(const ChunkID &) const;
+  friend unsigned int qHash(const ChunkID &);
+ protected:
+  int cx, cz;
+};
+
+inline ChunkID::ChunkID(int cx, int cz) : cx(cx), cz(cz) {
+}
+
+inline bool ChunkID::operator==(const ChunkID &other) const {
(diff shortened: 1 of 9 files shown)


chunkid.h
```diff
@@ -0,0 +1,26 @@
+#ifndef CHUNKID_H
+#define CHUNKID_H
+
+// ChunkID is the key used to identify entries in the Cache
+// Chunks are identified by their coordinates (CX,CZ) but a single key is needed to access a map like structure
+class ChunkID {
+ public:
+  ChunkID(int cx, int cz);
+  bool operator==(const ChunkID &) const;
+  friend unsigned int qHash(const ChunkID &);
```
madmaxoft: The previous declaration returned an `uint`; with bitwise operators that's the only safe way, some operations are undefined behavior on certain signed types. Are you still sure you need an `int`?

cre4ture: I replaced `uint` with `unsigned int` which should be the very same at the end.
take a look into `qglobal.h` and you find this definition of uint: `typedef unsigned int uint;`
I just found it more elegant to not have any dependency to qt here in the header anymore.
But it's rather a subtle difference considering that the most things in minutor make use of Qt stuff ;-). So if you see an other reason why you would prefer `uint`over `unsigned int` I'll happy to change it back - no problem.

madmaxoft: Right, I'm blind, completely missed the `unsigned` there. Sorry.

cre4ture:
I just did an other step and added it. Apperently this also updates the Pull Request automatically when pushed.

PR #215, opened 2020-03-28
search for blocks with certain id (like chests or moster spawner) and entities matching certain criterions like selling an enchanted book

cre4ture:
Full feature described in #199. 
I was initially quite pessimistic in my estimation about that it might take months for me to get this to master. I was able to cut away multiple larger changes that I did on my original feature branch but are actually not needed just for this feature.


cre4ture:
forgot to mention that it's preparation for issue #199

searchchunkswidget.cpp
```diff
@@ -0,0 +1,185 @@
...
+#include "ui_searchchunkswidget.h"
+
+#include "./chunkcache.h"
+#include "./chunkmath.h"
+#include "./range.h"
+#include "./rectangleinnertoouteriterator.h"
+
+#include <QVariant>
+#include <QTreeWidgetItem>
+#include <boost/noncopyable.hpp>
```
EtlamGit, line 11: Is it necessary to use BOOST here?
I do not have boost installed and therefore it does not compile anymore.
Uncommenting this line let it compile again.

mrkite: Seems unnecessary.  I removed the include.. and definitely agree that we should never use boost if we can help it.

cre4ture: Thanks for spotting and removing this. Seems to be a leftover that I didn't notice as I have boost installed. The boost::noncopyable is a very small class which at itself doesn't justify to add the dependency to boost - I agree.

PR #216, opened 2020-03-28
fix rare rendering bug and improve stability of chunk loading

cre4ture:
I stumbled across this by chance. I'm pretty sure that the `loaded = true;` needs to be at the very end if this function. 
In addition I extracted the function `Chunk::findHighestBlock` because the code uses a `return` for exiting from multiple nested loops. This would have made it impossible to have the `loaded = true;` at the very end.

I also recognized a random crash in `Chunk::load` caused by `memcpy` of invalid (zero-sized) IntArrayTags. This should at least fix the crash - still there might be an error in the logic of loading

### DIFF
--- a/chunk.cpp
+++ b/chunk.cpp
@@ -6,6 +6,19 @@
 #include "./flatteningconverter.h"
 #include "./blockidentifier.h"
 
+template<typename _ValueT>
+inline void* safeMemCpy(void* __dest, const std::vector<_ValueT>& __src, size_t __len)
+{
+  const size_t src_data_size = (sizeof(_ValueT) * __src.size());
+    if (__len > src_data_size)
+    {
+      __len = src_data_size; // this happens sometimes and I guess its then actually a bug in the load() implementation. But this way it at least doesn't crash randomly.
+    }
+
+    return memcpy(__dest, &__src[0], __len);
+}
+
+
 quint16 getBits(const unsigned char *data, int pos, int n) {
 //  quint16 result = 0;
   int arrIndex = pos/8;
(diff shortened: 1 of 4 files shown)


chunk.cpp
```diff
@@ -6,6 +6,19 @@
 #include "./flatteningconverter.h"
 #include "./blockidentifier.h"
 
+template<typename _ValueT>
+inline void* safeMemCpy(void* __dest, const std::vector<_ValueT>& __src, size_t __len)
```
madmaxoft: Why the underscores? And even double underscores?

cre4ture: Good point. There is no special reason. I used them because I copy-pasted the signature from original memcpy header definition. I fixed it now.

EtlamGit:
Where you ever able to debug one of these random crashes in `Chunk::load`?

Probably your `safeMemCpy` is a good place for a brake point or logging. I will try to debug that. Because at the end, the loading logic should be fixed.
In theory your extra check could also alter the performance. But I assume that is negligible.

All the changes are fine to me.

cre4ture:
> Where you ever able to debug one of these random crashes in Chunk::load
Yes I definitely had this cases where (as written in the description) the one of that arrays where actually zero length. I think it was typically with the biomes toIntArray, but it was also not reproducable all the time. Sadly I didn't note it down and it's some time ago already. But I can also put again a breakpoint into the safeMemCopy as you suggested and try to find a pattern in future.

cre4ture:
I have to correct myself: It is reproduceable at least with one of my worlds from 1.14.4.
It happens many times in this code part
`
      } else if ((this->version >= 1519)) {
        // raw copy Biome data
        safeMemCpy(this->biomes, biomes->toIntArray(), sizeof(int)*256);
}
`
I think it only happens when i display a region at the border of explored map area (where all these black rendered chunks are). I hope this helps you to find the issue.

mrkite:
Sorry about that, can you refactor this a bit so it merges cleanly?

PR #218, opened 2020-04-24
support for new BlockStates format comming with 20w17a

### LINKED ISSUE #217: Minecraft 1.16 pre-release 20w17a changed storage format
Mojang did some optimization for faster loading. This breaks loading of some Chunks.

EtlamGit:
fixes #217

### DIFF
--- a/chunk.cpp
+++ b/chunk.cpp
@@ -41,6 +41,7 @@ Chunk::~Chunk() {
         delete sections[i];
         sections[i] = NULL;
       }
+    loaded = false;
   }
 }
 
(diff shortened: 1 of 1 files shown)


EtlamGit:
@nnooney could you please review the new bit fiddling code I created (in second commit). For me everything renders fine, but I do not need byte swaps anymore.
I still had no time to compare with Mojang code, but I confident that they are doing it in a very similar way. Also the speed improvements they state for the new "optimized loading" in 20w17a is reasonable.

EtlamGit:
For everybody else:
It would be nice if someone can provide a test world with a section containing 4096 unique blocks. They have to be different at least in properties to create additional palette entries.

Other (high) numbers of different blocks be section are also welcome.

nnooney:
The code looks correct to me. When I was writing the original code, I was thinking about ordering the entire BlockStates section and then grabbing the right set of bits from a bit stream. Your code is much easier to reason about, and avoiding the memcpy and std::reverse definitely makes it more performant.

PR #235, opened 2020-08-28
build fully-functional app for macOS

Kolcha:
deploy Qt runtime and package .app into .dmg image making result
app complete and ready to be used by end-users.

unfortunately, archive created by GitHub Actions artifacts upload
step removes top level folder and archives only its content.
this is completely unacceptable, it is even not macOS app from fs
structure point of view despite on that is unusable without Qt...

running `macdeployqt` solves both problems: it integrates required
Qt runtime into .app bundle and also creates .dmg image (but later
it is archived into .zip, ugh..)

### DIFF
--- a/.github/workflows/c++qt-crossplattform.yml
+++ b/.github/workflows/c++qt-crossplattform.yml
@@ -30,11 +30,11 @@ jobs:
           - name: macos-5.9
             os: macos-latest
             qt: '5.9.9'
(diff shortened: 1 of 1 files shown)


Kolcha:
after this changes, macOS artifacts produced by CI are end-user usable app,
but without #234 only `macos-5.12` build produces usable app. see details in #234

PR #240, opened 2020-11-04
Villager poi visualisation

EtlamGit:
Comment before Merge!

Villagers have home and job locations (also called POI system).
Also PiglinBrutes use a home location to patrol around a given location.

These changes will draw small indication lines towards these locations.
I took simple blue for bed and red for work, as they differ well from typical block colors.
But I'm open for better color suggestions.
Also another location is possible to highlight: Meeting Point

Example with 4 farmers working in separate houses:
grafik

Still open is the dimension. All the locations are stored together with a dimension and should only be drawn when the current rendered dimension is the same.

### DIFF
--- a/entity.cpp
+++ b/entity.cpp
@@ -7,9 +7,9 @@
 Entity::Entity(const Point &positionInfo)
     : extraColor(QColor::fromRgb(0,255,0))
     , pos(positionInfo)
-{
-
-}
+    , hasExtraR(false)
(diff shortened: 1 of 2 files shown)


EtlamGit:
Same example as above, but the second Farmer has a meeting_point (and all are sleeping):
grafik
A trading hall with a bell gives meeting_points for all of them:
grafik

I'm not sure if we need meeting_point.

Issue #242, opened 2021-02-12
Issues with upcoming1.17


EtlamGit:
There are some (expected) issues with upcoming 1.17 release of Minecraft:
* Entities will no longer be stored in the same file like map data. An extra folder "entities" will be used instead. [->Minutor does not show the Entities]
* Build hight will change. Now something like Y-level -64 to 319 will be possible. This means 4 Sections will be added below and above current data. [-> Minutor will crash]

BlaCoiso:
I've tried loading a 1.17 (21w11a) world in the latest version, and it worked without crashing, however the sliders only allow viewing from 255 to 0 and the new blocks show up as purple (since there's no definition for them)

EtlamGit:
In my [development branch](https://github.com/EtlamGit/minutor/tree/extended_build_height) some first code to handle the new build height is available for testing.

It would be nice if someone else can do some profiling and verification.

BlaCoiso:
I was making a test world to test these extreme cases (height of 16, height of 2k, everything below 0, everything above 0, etc), but realized that minutor does not load custom dimensions (these are stored at `dimensions/namespace/dimension_name`, inside each of these folders there's the usual `data`, `entities`, `region` folders), so I'm unable to test unless I modify the overworld dimension.

EtlamGit:
You have custom dimensions with:
* all Y below 0
* maximum Y at 2K
???
That would be something completely unexpected for me. Especially the variable amount of sections does not fit current structure in Minutor. Therefore that will definitely not work. My target for now is to support new build height in vanilla 1.17.

You may attach your test world here. Then I have a chance to think about it.

BlaCoiso:
Here's the test world:
[dim_test.zip](https://github.com/mrkite/minutor/files/6258181/dim_test.zip)

Custom dimensions in this world:
max_middle (this is the one with maximum Y at 2k): height of 4048, y from -2032 to 2015, generates terrain but only inside overworld coords, column of wool blocks at 0,0
lower_smaller (this is one of them with all Y below 0): height of 128, y from -256 to -127, does not generate terrain, column of wool blocks at 0,0
min_down (this is another one with all Y below 0 and also smallest Y range): height of 16, y from -128 to -115, does not generate terrain, column of wool blocks at 0,0
upper_higher (this is the one with all Y above the usual): height of 1024, y from 448 to 1471, does not generate terrain, column of wool blocks at 0,0

These ones don't work and just crash the game when trying to generate terrain and/or structures
[...]

PR #243, opened 2021-02-12
load entities also from extra folder

EtlamGit:
With upcoming version 1.17 Entities are stored in a separated folder.
This pull request parses this folder for Entities  also.

Should solve part of #242

### DIFF
--- a/chunk.cpp
+++ b/chunk.cpp
@@ -159,6 +159,21 @@ void Chunk::load(const NBT &nbt) {
   loaded = true; // needs to be at the end!
 }
 
+void Chunk::loadEntities(const NBT &nbt) {
+  // parse Entities in extra folder (1.17+)
+  if (version >= 2681) {
+    if (nbt.has("Entities")) {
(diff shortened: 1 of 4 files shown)


EtlamGit:
Yes, I know that CI "Ubuntu 16.4 with default Qt 5.5" does not work anymore.

I opened an [issue](https://github.com/miurahr/aqtinstall/issues/216) at the [aqtinstall](https://github.com/miurahr/aqtinstall) project already.
If they can not solve it, we should drop this old Ubuntu. Ubuntu Xenial might be discontinued in April 2021 anyway.

EtlamGit:
After Mojang split 1.17 release in two parts, this should be the only needed change for 1.17.

Merge is save, we removed Ubuntu 16.04.

EtlamGit:
Definition update for 1.17 should be finished now!
Minecraft Release 1.17 is announced for next Tuesday.
This patch is mandatory to show Entities for 1.17.

PR #244, opened 2021-02-12
upgrade Qt version for Ubuntu 16.04

EtlamGit:
download link for old QT version was removed at qt.io

We could upgrade Qt to 5.9.5 (same as with Ubuntu 18.04) and our CI would work again.

I would prefer to remove Ubuntu 16.04 completely.
Just give your comments!


mrkite:
I definitely vote for removing 16.04 support.  Supporting 2 LTS versions is more than enough.

PR #262, opened 2021-11-15
fixed stuff for m1.18 extended build height

EtlamGit:
+ adapt JumpTo dialog
+ adapt Search: dialog, code and rendering

While looking into the "search code", I noticed different code style and much broken functionality.
+ adapted code style (therefore changed lot of files)
+ tried to fix some of the functionality
+ enhanced search dialog with pre-filled suggestions, auto scale and better readable results

But this topic is too big for the moment, I will open an issue after release 18 to collect question and ideas for search related stuff.


mrkite:
Failed to compile the searchblockplugin form on linux for some reason.

EtlamGit:
Yes, noticed & fixed that.
Very strange, as the change was already in d69b4cf, but that one reported to compile?

EtlamGit:
These are the last 1.18 related changes. No Block or definition update is needed this time.
I will try to do a small menu cleanup during next week, then we can do a 2.18.0 release just in time for Minecraft 1.18 release.

PR #263, opened 2021-11-18
restructure Overlay menu

EtlamGit:
+ flatten Overlay stuff into own menu
Beforehand it was quite tedious to enable/disable many structure overlays. You always had to open 2 sub menu levels where aiming for the second one requires much mouse movement.
Now all Overlays are assembled in an own menu at top level. So they are easier to reach, even as combination of keyboard (Alt+O) and mouse.

+ move all static menu creation into UI designer
All static part of the menu and actions creation is now moved to a from in UI Designer. It is easy to modify the menu there with drag and drop. Only dynamic part of the menu stay in the code: worlds to be loaded, player locations and Overlays

+ fix: clear structure overlay menu on world close
There was a bug, that when closing a world (loading another one), the Structure Overlays where not reset. So you get structures to select that do not exist. Now the Structure Overlay part of the menu is cleared correctly on world close.

**There is one point left:**
Position of "Manage Definitions". In my opinion it does not match the View menu, as it does not toggle any view modes. It should be located near something like: settings, preferences or update.
We have "Check for updates" in Help menu which is also not 100% expected. But also not completely uncommon.

Any suggestions for a better location?
(I would add a commit moving it.)

### DIFF
--- a/labelledseparator.cpp
+++ b/labelledseparator.cpp
@@ -1,0 +1,27 @@
+/** Copyright (c) 2021, EtlamGit */
+#include "labelledseparator.h"
+
+LabeledSeparator::LabeledSeparator(QWidget* parent)
+  : QWidgetAction(parent)
+{
+  line.setFrameShape(QFrame::HLine);
+  line.setFrameShadow(QFrame::Sunken);
+  line.setSizePolicy(QSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed));
+
+  layout.addWidget(&label);
+  layout.addWidget(&line);
+  widget.setLayout(&layout);
+  widget.setEnabled(false);
+  this->setDefaultWidget(&widget);
+}
+
+LabeledSeparator::LabeledSeparator(const QString& text, QWidget* parent)
+  : LabeledSeparator(parent)
+{
+  this->setText(text);
+}
+
+void LabeledSeparator::setText(const QString& text)
+{
+  label.setText(text);
+}
--- a/labelledseparator.h
+++ b/labelledseparator.h
@@ -1,0 +1,27 @@
+/** Copyright (c) 2021, EtlamGit */
+#ifndef LABELEDSEPARATOR_H
+#define LABELEDSEPARATOR_H
+
+#include <QWidgetAction>
+#include <QLabel>
+#include <QFrame>
+#include <QLayout>
+
+class LabeledSeparator: public QWidgetAction
+{
+  Q_OBJECT
+
+ public:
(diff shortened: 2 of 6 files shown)


EtlamGit:
grafik

EtlamGit:
@mrkite: any comment on the position of "Manage Definitions" menu item?

mrkite:
I agree it should be part of some settings/preferences menu.  One problem is the whole cross-platform thing.  On the mac, the help menu disappears entirely and that stuff gets stuffed into the app menu (which is where the preferences are typically found) but on windows, it'd all stay under Help which is not where I would look for preferences.

Maybe a Tools menu that has both check for updates and manage defintions?

EtlamGit:
Aha, did not know that with the Mac. But that means, that "Settings" and "Check for updates" are already moved into the app menu. So for Mac best position for all 3 items would be help menu.

I think all 3 belong together: "Settings" "Check for updates" "Manage definitions"

On Windows I also do not expect them at help, but e.g. Firefox has Help->About->Update.
We had it there in the past and moving "Settings" and "Check for updates" leaves only "About" in Help menu.

An own menu feels to much for stuff nobody is using regularly.
And I would expect other stuff in a Tools menu but do not find a better name.
I like these options:
* move all to Help
* move all to File

Or integrate the "Manage definitions" elements and an "Update" button into Settings.
But where to place Settings in that case?

(another option is to leave it as it is . . .)

mrkite:
I'd vote for move all to Help then, since that's what Qt expects.

madmaxoft:
Quite a few programs have Preferences in their Edit menu, that's an option I'd vote for.

EtlamGit:
Yes an Edit menu would be a very good place. But we do not have one now.

This would lead to the same situation as with Tools menu:
One extra menu with 3 items that are not used regularly and could also be combined into one single Settings item.

EtlamGit:
A "Force the Definition Update NOW" button in Settings looks like this:
grafik

madmaxoft:
I've just noticed that this PR removed the `Ctrl+D` keyboard shortcut for View -> Depth Shading, is that intentional, is there a reason for that, or can I add it back? I do tend to toggle this flag a lot, so it was a surprise to see it gone.

EtlamGit:
It is not intentional. At least I can not remember a good reason why I should have removed it.

just add into minutor.ui at lines 284++:
```
   <property name="shortcut">
    <string>Ctrl+D</string>
   </property>
```

madmaxoft:
Thx, re-added.

PR #265, opened 2021-11-24
more consistent keyboard navigation in menu

EtlamGit:
In some menus we had the same keyboard keys for multiple action items. E.g. "Open world" and "Open" had both "O".
Also some items did not have a keyboard shortcut.
Now keyboard navigation should be more consistent.


EtlamGit:
I will bump version number and prepare next release in the next days.

PR #268, opened 2021-12-18
Show spawners as overlay

### LINKED ISSUE #146: Visualize dungeons as structures
Despite not being recorded in the structures, it would be nice to have dungeons visualized similarly, so that they are easier to find.

This could be done as a heuristic in the chunk loader - if the loaded chunk contains a mob spawner, check the block underneath it, if it's a mossy cobblestone, add a dungeon entry in the structures array.

Alternatively, this could visualize all mob spawners, since they could be a valuable target.

EtlamGit:
This changes show spawners in "dungeons" as normal structure overlays. (fixes #146)

From saved file perspective they do not belong to structures, but from user perspective they are also some pre-generated structure found in the world.

It would be possible to parse chests the same way, but some refactoring is necessary before these information can be handled in Minutor.

The properties dialog was also modified:
* set title based on object we show
* disabled the "not used" line edit on top (to be reactivated later)

### DIFF
--- a/properties.cpp
+++ b/properties.cpp
@@ -28,15 +28,31 @@ void Properties::DisplayProperties(QVariant p) {
 
   // only support QVariantMap or QVariantHash at this level
   switch ((QMetaType::Type)p.type()) {
-    case QMetaType::QVariantMap:
-      treeCreator.ParseIterable(ui->propertyView->invisibleRootItem(), p.toMap());
+    case QMetaType::QVariantMap: {
+      // we got a QMap with all properties stored as [key]->value
+      auto data = p.toMap();
+      treeCreator.ParseIterable(ui->propertyView->invisibleRootItem(), data);
+      this->setWindowTitle(data.value("id", "Properties").toString());
       break;
-    case QMetaType::QVariantHash:
-      treeCreator.ParseIterable(ui->propertyView->invisibleRootItem(), p.toHash());
+    }
+    case QMetaType::QVariantHash: {
+      // we got a QHashMap with all properties stored as [key]->value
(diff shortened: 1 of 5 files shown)


properties.cpp
```diff
@@ -28,15 +28,31 @@ void Properties::DisplayProperties(QVariant p) {
 
   // only support QVariantMap or QVariantHash at this level
```
madmaxoft, line 29: This comment is now wrong

properties.ui
```diff
@@ -11,11 +11,15 @@
    </rect>
   </property>
   <property name="windowTitle">
-   <string>Dialog</string>
+   <string>Properties Dialog</string>
```
madmaxoft, line 14: Do we want the `Dialog` part at all? For me just `Properties` sounds better. (I assume this is the window title visible to the user, not some internal name.

madmaxoft (changes requested):

properties.ui
EtlamGit: This is the default title of the dialog.
In normal circumstances it should be overwritten by the ID (type of) whatever is shown in the dialog at that moment.

Yes, "Properties" is sufficient. We do not need to mention it is a dialog.

I plan to rework that dialog completely next year. There is some unused LineEdit on top, probably introduced to search in the properties. The plan is to try that out.

properties.cpp
EtlamGit: Yes and no.
The third option is a list which holds implicitly QVariantMap or QVariantHash.

But I did not completely figured out what is supported here also.
From what I have learned through debugging, only the first and third options are used in current code.

PR #271, opened 2021-12-25
Structure code with folders

EtlamGit:
move files that belong to feature/topic into own sub-folders

This is to have a more clean view on the main Minutor landing page at GitHub
and to get the sources organized in some way.

No functional change (except one quick fix to Search math by deleting a file).


EtlamGit:
Hmm, include path on Linux seem to be different ?!?

PR #275, opened 2022-01-09
mob spawn highlighting for water mobs

EtlamGit:
Hostile Drowned can spawn in water in Ocean or River Biomes.
This adds an additional spawn check for them.

In case we might miss-detect Ocean or River for future Biomes, a new JSON tag in vanilla_biomes.json is supported to tag that correctly.


EtlamGit:
Here I placed some light sources under water (Lighting + Mob spawning):
grafik

PR #281, opened 2022-01-23
fix rendering of only top half filled sections

### LINKED ISSUE #280: 2.18.0 not rendering some blocks at y=255 with 1.12.2 worlds
I have some issues with Minutor 2.18.0 not rendering blocks at y=255 in a 1.12.2 world.
Minutor 2.17.0 does not have this issue.

Here is a 1.12.2 region to reproduce (there is Obsidian at most areas of y=255)
[r.0.0-1.12.2.zip](https://github.com/mrkite/minutor/files/7920214/r.0.0-1.12.2.zip)

Map rendered with Minutor 2.17.0

Map rendered with Minutor 2.18.0

EtlamGit:
For old versions (before 1.13 "The Flattening") sections that contain only data in upper half are not rendered anymore.
Fix is just to check all Blocks in that Section to judge if it contains data (not lower half) . . .
( fixes #280 )

### DIFF
--- a/.github/workflows/cpp-qt-macos.yml
+++ b/.github/workflows/cpp-qt-macos.yml
@@ -1,4 +1,4 @@
-name: C++ Qt CI
+name: MacOS Qt CI
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


EtlamGit:
GitHub has changed windows-latest to "Windows Server 2022".

Issue #269, opened 2021-12-19
Support for more dimensions than default ones.


MajsterTynek:
Look into `20w14infinity` version and how saves work. 
This format may be used in the future or by mods.
+ add dimension parameter for commandline rendering

EtlamGit:
Work on custom dimensions is already ongoing:
* rendering is finished
* detection is discussed in #255 (and #242)
* some internal restructuring is necessary for proper support
* 20w14infinity is the Aprils Fools version, other more reasonable documentation is available

Yes you are right, a command line switch to select the dimension is also necessary.

EtlamGit:
@BlaCoiso & @MajsterTynek could you please test this new version from PullRequest #283 with some Custom Dimensions?

BlaCoiso:
I can try testing it, I also have a 20w14infinity world if that needs to be tested as well

EtlamGit:
20w14infinity was just an Aprils fool version. Implementation and format might have changed dramatically.
If it works, great. If not, we will not add support.

I implemented from documentation in Minecraft Wiki and tested with the last example world you have provided.

EtlamGit:
I removed the duplicate label, as now the only remaining part is: missing support on command line.

BlaCoiso:
https://github.com/mrkite/minutor/blob/085e5192335eeb9d442885fd56f363a95601bde1/worldinfo.cpp#L78-L90
There can be multiple dimensions using the same dimension type, or dimension types with different names than the dimension itself; dimension type specifies properties like y limits, skylight, etc while dimension specifies biomes, worldgen settings and which type to use
https://github.com/mrkite/minutor/blob/085e5192335eeb9d442885fd56f363a95601bde1/worldinfo.cpp#L69-L75
Datapacks can also be just a subfolder instead of a zip, I'm assuming for now support for subfolders rather than zip files is not a priority

Other than this, loading the custom dimensions worked just fine, although the Y pos label is getting cropped (Y > 999 or Y < -99)

EtlamGit:
So I have to somehow  parse `dimension` and `dimension_type`. I will think about that.

Cropped Y is known, I hope that nobody (except you) will try something that large.
But it should be solvable with correct UI definition.

MajsterTynek:
Dimension parameter should be possible to pair with `--jump`.
For example, I would want to make a script that opens 
the current world at a current location I am in the game,
but for now, it defaults to the overworld and starts minimized. ¯\\\_(ツ)\_/¯

[...]

Issue #255, opened 2021-10-08
Issues with upcoming 1.18 (extended build height)


AndrewIsOffline:
If an area is saved in a world running 21w39a or up, the area won't render properly in the viewer. The biome will show up as either unknown or ocean, and the blocks will show up as minecraft:air. This is best illustrated if the player loads in far away from world spawn:
![image](https://user-images.githubusercontent.com/6992926/136543363-af83c2c5-66dd-4617-bf8c-83c2962cf556.png)
This might have to do with the new negative y coordinates.

EtlamGit:
Mojang did change the storage format again . . . 

* Now Biomes are not stored anymore per Chunk, put per Section. They also have a Palette that has to be decoded.
* BlockState is still stored in the same manner as before (with Palette), but the NBT keys have changed name and structure.

@AndrewIsOffline did you check, that the version number you provided is exactly the very first snapshot showing this behavior?

EtlamGit:
The change notes for 21w39a indicate the modified Biomes, so it should be ok for the moment.

EtlamGit:
Some intermediate status:
* I'm able to render latest snapshot again, will publish the changes soon
* more Biome data (now per Section) and more Sections will lead to about double of memory usage, so we will only be able to show half sized maps compared to now

EtlamGit:
Everything is working for 1.18, the release is out and finally also Minecraft released 1.18 today.

@BlaCoiso: As I did extended build height implementation twice during the this year, the second time I prepared everything to render arbitrary world height. So in principle it will work with your strange dimensions. The only step missing is to detect the availability of custom dimensions.

Could you please provide some test worlds with your custom dimensions?
* I do not have the time to prepare these custom worlds.
* But I need them to figure out how to detect a custom dimension and how to detect build height for that dimension.

BlaCoiso:
I've found an issue with the search for block feature where it ignores blocks at Y=0; about the custom dimensions, I'll look at the current format for specifying them and try to generate them again, I'll send the new world with those dimensions once I have them ready

EtlamGit:
continue in #269

PR #283, opened 2022-01-30
add WorldInfo as central information collection point

EtlamGit:
To support some features we need a place to store information that is common for the complete world. Up to know we only have Definitions which are common to all worlds and Chunk data that is common only to a single Chunk. Some exceptions were solved with extra variables.
(Example features are: World name, Spawn coordinates, last saved data version, Custom Dimensions and Local Difficulty Highlight.)

Here I started to add the class WorldInfo to hold all these world common information.
Then I migrated all the extra complicated stuff, what went smooth.
And tested some custom dimensions, what was working faster then expected.

@mrkite: adding the Custom Dimensions to the menu feels still a bit ugly. Do you have some better idea for that part?

It is the first step to solve #269, #255 and #242.
Here a first quick test worked with custom Dimensions. Maybe it was all, but I think we will need some deeper testing.

### DIFF
--- a/definitions/vanilla_dims.json
+++ b/definitions/vanilla_dims.json
@@ -1,7 +1,7 @@
 {
   "name": "Vanilla",
   "type": "dimension",
-  "version": "1.18.0",
+  "version": "1.18.1",
   "data": [
     {
       "name": "Overworld",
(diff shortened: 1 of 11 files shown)


mrkite:
I don't have any better ideas for the custom dimensions..

EtlamGit:
I have the tendency to move all world related Menu stuff into WorldInfo.
All other Identifers only provide information when asked. DimensionIdentifer is the only one that has more functionality.

But that is a maintenance task and can be done later.

PR #286, opened 2022-02-10
Store zoom level as a double

### LINKED ISSUE #224: Zooming the map by scrolling does not work as expected on Touchpad
Hi,

I have a touchpad that uses two finger inputs to scroll. Scrolling works in every other program. In Minutor, attempting to scroll will either do nothing, zoom *in* to a random zoom level, or zoom *out* to a random zoom level (including randomly jerking in and out across multiple zoom levels even if I am scrolling continuously on one direction on the touchpad). 

This is on KDE Plasma (latest released version) on Linux. I believe my touchpad is using the xf86-input-synaptics driver.

Things I have tried:

 * Set up a script with `xdotool` to send one "scroll up" input every five seconds. This *works*. Minutor goes up one zoom level every five seconds.

 * Checked the inputs being sent to programs by the mouse driver with `xev`. If I scroll extremely slowly, I can get the touchpad to only send two scroll events at a time. I can't get it down to one. Note that scrolling extremely slowly does not seem to solve the problem with scrolling in Minutor.

If you don't immediately know how to fix this problem, or fixing it would create other issues, I think it would be nice to have some kind of keyboard input for scrolling as a around, and possible a zoom slider on the right of the screen as well, like the "layers" slider.

afontenot:
This changes the way that the current zoom level is stored by the
program. Previously, only the integer index of the zoom level was
stored. This leads to an issue: when you have many scroll events
firing every second, each individual scroll event will only have
a tiny delta. Most of these deltas will not be large enough to
change the zoom level, but they should build up over time. The
previous implementation lost these deltas because of rounding.

With this change, we store the zoom level as a double, and only
convert this to an index (by rounding) when this is needed to
change the scaling factor.

I also fixed a couple of tiny issues:

1. changed delta -> angleDelta to fix QT deprecation warnings.

2. preventing having to call redraw() for every scroll event

Fixes #224

### DIFF
--- a/mapview.cpp
+++ b/mapview.cpp
@@ -2,6 +2,7 @@
 #include <QPainter>
 #include <QResizeEvent>
 #include <QMessageBox>
+#include <cmath>
 #include <assert.h>
 
 #include "mapview.h"
@@ -140,7 +141,7 @@ void MapView::clearCache() {
 void MapView::adjustZoom(double steps, bool allowZoomOut)
 {
   // get current zoom level as an index, rounded to nearest int
-  int oldZoomIndex = zoomLevel + 0.5;
+  int oldZoomIndex = (int)(floor(zoomLevel + 0.5));
   zoomLevel += steps;
 
   // use Fibonacci numbers to get natural zoom behaviour
@@ -151,7 +152,7 @@ void MapView::adjustZoom(double steps, bool allowZoomOut)
   if (zoomLevel < zoomMin) zoomLevel = zoomMin;
   if (zoomLevel > zoomMax) zoomLevel = zoomMax;
 
-  int zoomIndex = zoomLevel + 0.5;
+  int zoomIndex = (int)(floor(zoomLevel + 0.5 ));
 
   // check whether zoomIndex has changed since last adjustZoom call
   // don't return early if steps == 0, this is used for initialization
--- a/mapview.cpp
+++ b/mapview.cpp
@@ -16,7 +16,7 @@ MapView::MapView(QWidget *parent)
   : QWidget(parent)
   , depth(255)
   , scale(1)      // overworld coordinate mapping
-  , zoomIndex(0)  // 1:1
+  , zoomLevel(0)  // 1:1
   , cache(ChunkCache::Instance())
 {
   adjustZoom(0, false);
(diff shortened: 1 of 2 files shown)


EtlamGit:
Regarding the build error on Ubuntu 20.04:
It is because the current ubuntu-latest image has an outdated apt cache
If we wait a week this should solve itself . . . but I added an apt-get update to the build flow.

So we might just ignore it for the moment, or @afontenot has to rebase this branch.

afontenot:
@EtlamGit I've rebased on top of your fix for the CI, and also made a small change to allow merging cleanly with my fix for the issue with the depth slider: https://github.com/mrkite/minutor/pull/288

EtlamGit:
Still have to check the rounding behavior. (Probably working in this case.)
I always get nervous when I seen `(int) some_real + 0.5`
Especially in my normal work topics, but also here in Minutor the search dialog was broken because of this style of rounding.

So I may suggest to make that rounding part more explicit.

afontenot:
>So I may suggest to make that rounding part more explicit.

I figured it was alright in this case because the values are guaranteed to be positive (and range checks are required anyway). If you like, I can change it to use `round()` or use an explicit cast to make the assumption more obvious.

EtlamGit:
https://www.onlinegdb.com/online_c++_compiler

```
#include <iostream>
#include <cmath>

int main()
{
    std::cout << "Show rounding\n";
    for (double d=-3.0; d<=3.0; d+=0.25) {
        printf("%+.2f -> %+2d %+2d %+2d %+2d %+2d\n", d,
                int(d+0.5),
                int(round(d)),
                int(trunc(d+0.5)),
                int(floor(d+0.5)),
                int(ceil(d-0.5))
              );  
    }

    return 0;
}
```
First two variants generate strange results for negative numbers. The group of identical integer numbers (same index) is different around zero. If you use +0.5 with truncation you get index0 7 instead of 4 times. And with round() you get index0 3 instead of 4 times.
What we need is to round in the same direction on both sides of zero.

EtlamGit:
And ZoomIndex can get negative.

Have you tried to use a modifier key when zoomed out to 1:1? You can zoom out further!

EtlamGit:
I like `(int)(floor(double_value+0.5))` the most.

afontenot:
> And ZoomIndex can get negative.
> 
> Have you tried to use a modifier key when zoomed out to 1:1? You can zoom out further!

Good catch. I didn't know that feature existed and didn't look at that part of the code closely when I was fixing this. I think `floor` is probably the wrong way to go, though. Unless I make the code more complicated, you won't be able to get to the highest zoom level as easily. You will need to exceed the maximum value. Likewise, there wouldn't be as much space around 0: only 0-0.5 would result a zoom level of zero.

Any reason not to just use `std::lround`?

EtlamGit:
`(int)(floor(double_value + 0.5 ))` allows to reach the maximum value exactly 0.5 digits before reaching it. That is the same behavior like your code or round() regarding positive numbers.

lround() has the same (wrong) behavior like round(): "rounding halfway cases away from zero"

Just add it to the online code example and try.

afontenot:
> rounding halfway cases away from zero

That doesn't seem like too much of an issue: since we're dealing with floating point values (and adding in a bunch of scroll event deltas with a lot of decimal places) hitting an exact `0.5` is so rare as to not really be worth worrying about. Using the `std::` functions seems like a cleaner approach to me, but I'll go with `(int)(floor(double_value + 0.5 ))` if you like.

EtlamGit:
`std::floor()` is also `std::`

I have no clue what MacOS is sending there as mouse events. On my PC I always get increments of 120, so I will never have non integer numbers here.
But I assume MacOS will send integer increments in range 1..120 depending on the HighRes factor they use. Where 60 is for us 60/120 = 0.5 So it might hit the half quite often, in case HighRes is 240pdi it will be every second one, in case of 4800 dpi only every 40th.

EtlamGit:
In real live I get paid to discuss atomic small deviations resulting from math functions like round().
Maybe therefore I'm more picky with that topic.

afontenot:
Fair points, I've taken your suggestion and committed it.

> On my PC I always get increments of 120, so I will never have non integer numbers here.

I hadn't considered that there were PCs that were sending values that high. A comfortable scrolling speed for me in Minutor gets me values of 10-20. This raises another possible issue though (not with my PR): suppose there *is* some system out there that only returns scroll delta values >=240. In this case there will be no way for the user to scroll to individual zoom levels: they will always skip over one level even with the slowest scrolling they can achieve with their device.

afontenot:
@mrkite If you want a cleaner history let me know and I'll squash the changes in my PRs into a single commit per PR.

EtlamGit:
Yes, in case such computers exist we will get into trouble. In case I understand the Qt documentation correctly, the expected value is 120 and smaller in case of HighRes devices (most probably MacOS).

My hope is that this will never occur . . . or we will get an Issue opened by a frustrated user.

EtlamGit (approved):

PR #288, opened 2022-02-12
Use a double to store state of LabelledSlider

### LINKED ISSUE #287: Scrolling with modifier key does not move depth slider as expected on MacOS
On MacBook Pro with Mac OS Monterey, scrolling while holding a modifier key does not move the depth slider as expected.

The depth slider itself works fine.

Fn plus left and right cursor keys moves the depth slider in 1-block increments as expected.

Using shift as a modifier key, I cannot move the depth slider using my mouse scroll wheel. Using the touch pad, I can move the depth slider in large increments by holding shift and performing aggressive flicking movements, but not the normal two-finger scroll gesture. This seems similar to the previous issue #224 with scrolling to zoom as described by @rjp [here](https://github.com/mrkite/minutor/issues/224#issuecomment-992725667).

afontenot:
The LabelledSlider is used to change the depth setting. Some
touchpads (e.g. on macOS) can send dozens or hundreds of scroll
events per second, each with a very small delta. Unless we keep
track of the "true" state of the slider, these deltas will be lost
when they are rounded into the integer depth setting.

Fixes #287

### DIFF
--- a/labelledslider.cpp
+++ b/labelledslider.cpp
@@ -38,15 +38,14 @@ int LabelledSlider::value() const {
 // public slot
 void LabelledSlider::setValue(double v) {
   preciseValue = v;
+  if (preciseValue < slider->minimum()) preciseValue = slider->minimum();
+  if (preciseValue > slider->maximum()) preciseValue = slider->maximum();
   slider->setValue((int)(v + 0.5));
 }
 
 // public slot
 void LabelledSlider::changeValue(double v) {
-  preciseValue += v;
-  if (preciseValue < slider->minimum()) preciseValue = slider->minimum();
-  if (preciseValue > slider->maximum()) preciseValue = slider->maximum();
-  slider->setValue(preciseValue);
+  this->setValue(preciseValue + v);
 }
 
 // public slot
--- a/labelledslider.cpp
+++ b/labelledslider.cpp
@@ -1,5 +1,6 @@
 /** Copyright (c) 2013, Sean Kasun */
 #include <QVBoxLayout>
+#include <cmath>
 #include "labelledslider.h"
 
 LabelledSlider::LabelledSlider(Qt::Orientation orientation, QWidget *parent) :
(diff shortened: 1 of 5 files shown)


labelledslider.cpp
```diff
@@ -36,13 +36,17 @@ int LabelledSlider::value() const {
 }
 
 // public slot
-void LabelledSlider::setValue(int v) {
-  slider->setValue(v);
+void LabelledSlider::setValue(double v) {
+  preciseValue = v;
+  slider->setValue((int)(v + 0.5));
```
EtlamGit, line 41: I like this more that the other places where rounding is done.
Here an explicit cast to int is visible for a reviewer, so intended rounding could be expected.

labelledslider.cpp
```diff
@@ -36,13 +36,17 @@ int LabelledSlider::value() const {
...
 }
 
 // public slot
-void LabelledSlider::changeValue(int v) {
-  slider->setValue(slider->value() + v);
+void LabelledSlider::changeValue(double v) {
+  preciseValue += v;
+  if (preciseValue < slider->minimum()) preciseValue = slider->minimum();
```
EtlamGit, line 47: maybe we should move that range check to `LabeledSlider::setValue(double v)`
otherwise it would be possible to write wrong values to preciseValue with that mehtod

afontenot: Sounds fine to me, I'll change it to do that if you like. But `LabeledSlider::setValue` didn't have that check before, however, and arguably it doesn't need it because it's just calling setValue on the slider and QSliders inherently limit their values to the minimum and maximum. https://doc.qt.io/qt-5/qabstractslider.html#value-prop

afontenot: Scratch that, I see what you mean. We have to set `preciseValue` there. I'll rework it.

afontenot: Fixed.

afontenot:
@EtlamGit I don't know whether you like merging multiple work-in-progress commits or not. If you like I'll move my changes into a single commit and force push before merging so you get a cleaner history.

EtlamGit:
I always let mrkite do the merge. That ensures that he has also seen and reviewed all new code.

EtlamGit (approved):

EtlamGit (approved):

labelledslider.cpp
madmaxoft: Is there a specific reason for avoiding C++style cast here?
`slider->setValue(static_cast<int>(v + 0.5));`

EtlamGit: slider->setValue(static_cast<int>floor(v + 0.5));

I just realized, that the DepthSlider can now be also negative . . .

EtlamGit (approved):

PR #295, opened 2022-03-08
When zooming with pointing device, zoom into cursor location

### LINKED ISSUE #294: Zooming by scrolling should center the zoom on the cursor location
Currently, no matter how you initiate a zoom, the zoom takes you towards or away from the center of the current view. This is undesirable when using a mouse / cursor, because you tend to be poking around with it looking at different things (e.g. by hovering), and start a zoom when you want to look at the thing you're hovering over more closely.

The desirable behavior is to keep the pixel that the cursor is currently touching still, and zoom everything else around it as before.

This would mirror the behavior of just about every mapping application in the world, try [OpenStreetMap](https://www.openstreetmap.org) as an example of what I mean.

I think the way to go here in terms of implementation would be to add `bool followCursor` to `adjustZoom`. Non-scroll wheel based zooms would always call this with `false`, scroll wheel zooms would call it with `true` **or** optionally call it with the value of a setting, if it's desired to make this behavior configurable. Doing the zoom would be easy, you just recenter the map as appropriate right before the `redraw()`.

Thoughts? If this would get merged, I might try working on it if no one else wants to do it.

afontenot:
Before this commit, zooming always zooms into the center of the
view. After this commit, we zoom in such a way that the point under
the cursor represents the same location on the map before and after
the zoom. The imitates the behavior of most online mapping tools,
e.g. OpenStreetMaps.

The new zooming approach only applies when using the scroll wheel,
not when zooming with the keyboard. It is also dependent on a
QSetting. If desired, a future commit could expose this setting in
the GUI, making it possible to disable the new zoom behavior.

Fixes #294.

### DIFF
--- a/mapview.cpp
+++ b/mapview.cpp
@@ -20,7 +20,7 @@ MapView::MapView(QWidget *parent)
   , zoomLevel(0)  // 1:1
   , cache(ChunkCache::Instance())
 {
-  adjustZoom(0, false);
+  adjustZoom(0, false, false);
   connect(&cache, SIGNAL(chunkLoaded(int, int)),
           this,   SLOT  (chunkUpdated(int, int)));
   connect(&cache, SIGNAL(structureFound(QSharedPointer<GeneratedStructure>)),
@@ -138,8 +138,11 @@ void MapView::clearCache() {
   redraw();
 }
 
-void MapView::adjustZoom(double steps, bool allowZoomOut)
+void MapView::adjustZoom(double steps, bool allowZoomOut, bool cursorSource)
 {
+  // save old zoom value for panning to cursor
+  double oldZoom = zoom;
+
   // get current zoom level as an index, rounded to nearest int
   int oldZoomIndex = (int)(floor(zoomLevel + 0.5));
   zoomLevel += steps;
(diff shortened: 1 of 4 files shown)


afontenot:
This turned out to be easier and more clean than I had expected. The only significant change required outside the `adjustZoom` function is to always keep track of the last seen cursor position on the map. This has no performance impact, given that we already have a callback for cursor events. Changing the zoom behavior requires calculating a new map center, but this is simple enough that I wasn't able to see any performance impact from having it enabled.

EtlamGit:
The [QWheelEvent ](https://doc.qt.io/qt-5/qwheelevent.html#position) has also information regarding the current cursor position. Maybe we can use that information and skip own tracking.
I will try to test that this evening.

afontenot:
It does, but we're already tracking the cursor on the MapView widget, see https://github.com/mrkite/minutor/blob/master/mapview.cpp#L28 We're already running a function for every cursor movement event, the only change I've made is having the function update the stored values every time so that they're available to the `adjustZoom` method.

That said, if you prefer only updating when there's a scroll event, I've verified that it does seem to work. Here's a branch with that change instead: https://github.com/afontenot/minutor/tree/zoom-to-cursor-alt

If you want, I'll close this PR and open one for that branch, or just force-push this branch.

EtlamGit (approved):

afontenot:
@EtlamGit since this has been merged, you can either move the other global variable yourself or I'll do another PR for it if you like.

Incidentally one useful thing you could do would be to add a `TODO` file in the root directory of the project containing all the desired improvements you mentioned here: https://github.com/mrkite/minutor/issues/294#issuecomment-1060821045

As of this commit, every problem I've had with the program has been fixed, but it might be that there are some features I'd like that I haven't even thought of, and a `TODO` would help others contribute to the project. It's probably easier than adding issues for all of them, though that approach would also be valid. Existing issues could be linked in the `TODO`.

PR #306, opened 2022-06-07
fix wrong qHash0 names

### LINKED ISSUE #305: [Windows] Error loading Block definition: minecraft:air
Always get the following error on launch:

Error loading Block definition: minecraft:air
Failed to add Block from definition file, as it might be a duplicate or generated the same hash as an already existing Block.

Only option is 'Cancel' which just opens the same message again (can only close with Task Manager).

I performed Windows updates, updated Visual C++ Redistributable, updated Java, reinstalled Minecraft and tried on another Windows 10 machine (the Linux version did end up working though).

EtlamGit:
Some time ago Qt changed the way qHash is used, so we migrated to use 42 as seed. Here we try to detect and remove some more of the old names.

will fix #305

### DIFF
--- a/identifier/definitionmanager.cpp
+++ b/identifier/definitionmanager.cpp
@@ -120,12 +120,6 @@ void DefinitionManager::checkAndRepair()
     }
   }
 
-  // repair when definition is in settings, but file is missing
-  for (const auto& def: known_packs) {
-    if (!QFile::exists(def.toString()))
-      known_packs.removeOne(def.toString());
-  }
-
   // copy over built-in definitions if necessary
   QDirIterator build_in(":/definitions", QDir::Files | QDir::Readable);
   while (build_in.hasNext()) {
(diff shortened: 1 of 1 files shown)


identifier/definitionmanager.cpp
```diff
@@ -258,6 +266,13 @@ void DefinitionManager::installJson(QString path, bool overwrite,
   QString key = def->at("name")->asString() + def->at("type")->asString();
   QString exeversion = def->at("version")->asString();
 
+  // check if intermediate qHash(..,0) name is present
+  QString dest0 = destdir + "/" + QString("%1").arg(qHash(key, 0)) + ".json";
+  if (QFile::exists(dest0)) {
+    QFile file (dest0);
+    file.remove();
+  }
+
```
madmaxoft, line 275: I think this should be more documented. At least a link to the issue.

EtlamGit: My decision was: "It is save to delete definition files"
In the past we had already a list with hard coded hashed names we always delete, this is extending this list.

Issue #246, opened 2021-03-24
Clearing chunk cache takes ages


madmaxoft:
As of commit 4ff205387562afc44f92bedf29fecea1788a7b5f, clearing the chunk cache (i. e. refreshing or reloading the map) takes an awful lot of time - long minutes (!) even in a Release build. It seems that most of the time is spent deleting the per-section palettes, mostly the `properties` member.

Tested on a self-compiled Win64 binary built with Qt 5.15.1.

EtlamGit:
I experienced this behavior for a much longer time. But only for Debug builds!

The mentioned commit only removes Ubuntu 16.04 support from the build chain. (No code change.)
So the problem is older and as mentioned probably in the Chunk destructor.

madmaxoft:
I did some digging around and finally found out the culprit is not Minutor itself, but the QtCreator's debugger. When running under the debugger, the `QMap` and `QString` operations take about 5 - **60** times longer than when running in a standalone executable. For me, this makes even Release builds barely usable under QtCreator, reloading after displaying just a screenful of a world takes several minutes.

I even made a simple test project to measure the `QMap<QString>` performance against `std::map<QString>`, `std::map<std::string>`, `QMap<std::string>`, `std::undordered_map<QString>` and `std::unordered_map<std::string>`. The test fills 50k instances of `map` with 30 items (9-char key, 9-char value) each. The results below are from a MSVC 2019 Release build on Win8.1, standalone executable:
```
QMapQString: Filling  the containers took 686 msec
QMapQString: Clearing the containers took 220 msec
StdMapQString: Filling  the containers took 507 msec
StdMapQString: Clearing the containers took 211 msec
StdMapStdString: Filling  the containers took 373 msec
StdMapStdString: Clearing the containers took 186 msec
QMapStdString: Filling  the containers took 493 msec
QMapStdString: Clearing the containers took 139 msec
StdUnorderedMapQString: Filling  the containers took 447 msec
StdUnorderedMapQString: Clearing the containers took 190 msec
StdUnorderedMapStdString: Filling  the containers took 336 msec
StdUnorderedMapStdString: Clearing the containers took 147 msec
```
[...]

PR #307, opened 2022-06-09
performance optimization: first steps

Kolcha:
these 2 commits improve just one function (see its description for details), but the result is impressive - the function became twice faster. and this noticeably affected the whole rendering.

before changes:
Screenshot_20220610_010132
after this changes:
Screenshot_20220610_010541

and this is what I noticed just looked over the sources... I will look more and likely find more similar prominent places that can be easily optimized. this is just the first step.

### DIFF
--- a/identifier/blockidentifier.cpp
+++ b/identifier/blockidentifier.cpp
@@ -110,9 +110,9 @@ BlockIdentifier& BlockIdentifier::Instance() {
 }
 
 BlockInfo &BlockIdentifier::getBlockInfo(uint hid) {
-  if (blocks.contains(hid)) {
-    return *blocks[hid];
-  }
+  auto iter = blocks.find(hid);
+  if (iter != blocks.end())
+    return *iter.value();
   // no blocks at all found.. dammit
   return unknownBlock;
 }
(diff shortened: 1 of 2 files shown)


EtlamGit:
Yes, for larger number of elements a QHash should be much faster than a QMap. And not searching twice is obviously faster.

Some time ago we measured performance of different containers (#246).
The outcome was, that Qt containers are in some cases slower than C++ std:: variants.
And at many locations there is no strict need to use the Qt containers. It is just cleaner to have it less mixed.
But there might be some performance to gain from that also.

My personal experience is that I do not suffer from rendering performance. I always struggle with file-IO performance (world is stored on network drive). Therefore it is one of my goals to improve file reading, rendering speed was neglected during the last years.

Kolcha:
first of all, I have no plans to achieve the maximum possible performance or smooth rendering (I'm also fine with delays), I just noticed the "bad place", and think such obvious things must be fixed/optimized, especially when it costs nothing (just a few lines must be changed with no consequences).

> The outcome was, that Qt containers are in some cases slower than C++ std:: variants.
> And at many locations there is no strict need to use the Qt containers. It is just cleaner to have it less mixed.

I completely agree, code is written in Qt - let's use Qt containers unless there is a strict reason do not so. and those Qt containers may be slower it is expected for me - Qt metaobject system definitely leaves some footprint on performance.

PR #313, opened 2022-06-11
fixes for issues detected by static code analyzer, worth fixing

Kolcha:
detected by static code analyzer, worth fixing

### DIFF
--- a/identifier/definitionmanager.cpp
+++ b/identifier/definitionmanager.cpp
@@ -437,7 +437,6 @@ void DefinitionManager::loadDefinition(QString path) {
     d.path = path;
     d.update = def->at("update")->asString();
     QString type = def->at("type")->asString();
-    QString key = d.name + type;
     d.enabled = true;  // should look this up
     if (type == "block") {
       d.id = flatteningConverter.addDefinitions(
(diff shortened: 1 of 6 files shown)


EtlamGit:
All your changes are correct.
The reason I left all these "unused" parameters/variables in **the search related code** is to have some kind of remainder for me that major cleanup is necessary. There is much non working or never finished code and also some of the functionality is broken with the new storage format. I tried to fix the most important stuff during the 1.18 update, but there is much more cleanup left. It is on my list of things that has to be improved during the next years.
So it is just a choice if we want to have clean code now or some nasty warnings that will force me to look into that code eventually.

As you like clean code, maybe you can place some "todo" comments each time you remove warnings there?
Then I can try to use these marks as remainder.

Kolcha:
> There is much non working or never finished code and also some of the functionality is broken with the new storage format. I tried to fix the most important stuff during the 1.18 update

didn't know that... sorry



> As you like clean code, maybe you can place some "todo" comments each time you remove warnings there?

placed 'TODO(search)' in the functions where unused parameters were (and silenced using Q_UNUSED)

EtlamGit (approved):

EtlamGit:
You could not know it, as I did not communicate it. So no problem at all.

I really appreciate, that someone is trying to improve code quality!

PR #318, opened 2022-07-02
drop custom JSON parser, use Qt's one instead

Kolcha:
since 5.0 Qt provides JSON parser with very similar interface to this custom one. so it is no need to keep custom parser anymore.

"smoke test" passed - tested with 1.8.9, 1.12.2, 1.15.2, 1.19, so very likely everything is fine.


EtlamGit:
From a first look the interface seems really very similar. And you are right: it is better to use tested Qt code instead to try maintaining our own implementation.

One thing I noticed, but did not have time to look deeper:
You changed several function parameters from pointer to direct object. I fear, that we copy lot of data each time. This might not be a big performance issue, as it only done once at startup. But maybe there is a way to use references and some "const" stuff at these places.

Kolcha:
it is safe to copy Qt "heavy" objects - they are "implicitly shared" and do the actual "deep copy" only on modification.
most Qt examples also don't care about passing objects by reference due to this copy-on-write logic (even Qt doesn't take advantage of C++ move semantics in most cases, at least for Qt 5).

so, no worries here.

EtlamGit:
That was the answer I wanted to hear ;-)

PR #324, opened 2022-07-30
DefinitionManager code cleanup

Kolcha:
- eliminate QVariant -> QString conversions (not breaking change, compatible with existing data)
- do not create unnecessary QDir and QFile objects to perform one operation

### DIFF
--- a/identifier/definitionmanager.cpp
+++ b/identifier/definitionmanager.cpp
@@ -79,9 +79,9 @@ DefinitionManager::DefinitionManager(QWidget *parent) :
 
   // we load the definitions in backwards order for priority
   QSettings settings;
-  sorted = settings.value("packs").toList();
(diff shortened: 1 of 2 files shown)


EtlamGit (approved):

EtlamGit (commented):
Did you test what is happening when someone forcefully replaces Minutor->packs by an incompatible data type (e.g. REG_DWORD)?
This probably caused issues before, so question is just if we can handle that.

Kolcha:
don't have Windows, so tested on Linux - just replaced human-readable comma separated list with some binary data (serialized list of ints - `@Variant(\0\0\0\x7f\0\0\0\vQList<int>\0)`). in such case just an empty list is returned (and some warning is produced in terminal). so, at least this doesn't cause the crash and there is the way to detect such case - empty list is returned (this can be considered as "missing key")

PR #328, opened 2022-08-24
return -1 for invalid biome offset in any case

Kolcha:
in case of invalid biome offset previously -1 was returned only
for Debug builds, causing undefined behavior for Release builds.

this is relevant only for Minecraft < 1.18, changed this to the
same as for 1.18+ code - return -1 in invalid case, and print
warning only for Debug builds.

### DIFF
--- a/chunk.cpp
+++ b/chunk.cpp
@@ -127,12 +127,12 @@ int Chunk::getBiomeID(int x, int y, int z) const {
     offset = x + 16*z;
   }
 
-  #if defined(DEBUG) || defined(_DEBUG) || defined(QT_DEBUG)
   if ((offset < 0) || ((unsigned long long)(offset) > sizeof(this->biomes)/sizeof(this->biomes[0]))) {
+    #if defined(DEBUG) || defined(_DEBUG) || defined(QT_DEBUG)
     qWarning() << "Biome index out of range!";
+    #endif
     return -1;
   }
-  #endif
   return this->biomes[offset];
 }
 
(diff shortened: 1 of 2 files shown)


EtlamGit:
I think we have to extend this fix:
When this function does return a -1 this value will get into one of these methods:
```
const BiomeInfo &BiomeIdentifier::getBiomeByChunk(qint32 id) const
const BiomeInfo &BiomeIdentifier::getBiomeBySection(qint32 id) const 
```
And these are not expecting negative numbers jet (at least the second one).

Kolcha:
```
const BiomeInfo &BiomeIdentifier::getBiomeByChunk(qint32 id) const
```
is safe - it will return `unknownBiome` object, which can be used (and is used these days)

```
const BiomeInfo &BiomeIdentifier::getBiomeBySection(qint32 id) const 
```
is updated to be prepared to invalid indexes (more correctly slightly extended to handle negative values)

Kolcha:
you can test these changes using [this world](https://www.dropbox.com/s/9800j7304erw54l/holed.zip?dl=1). it has "holes" (created in Creative mode) at (0, 0) in Overworld and Nether dimensions, hovering over them triggers "Section not found for Biome lookup!" message and as result -1 is passed to `getBiomeBySection()`. world is created by Minecraft 1.19.1

this is maybe unusual case, but initial issue was also detected using some "very strange" worlds. it worth fixing as part of error handling and general reliability improvement.

PR #341, opened 2022-11-02
use Biome color for empty Blocks in single layer

EtlamGit:
For normal rendering we use Biome color of the "highest" Block at each XZ position.

This change adds a new mode (when Biome Color Overlay and Single Layer are both selected):
Then the Biome color is derived from the Block at current Y depth, even when the Block is air.

In case you plan farms or builds in caves or skyblock worlds it is crucial to know the Biome for each Block even at specific depth layer (Biome is height dependent). With this new mode you can analyze Biome layer by layer.


EtlamGit:
Bug report for aqt is already filed.

EtlamGit:
Now the workflow is running again . . . very strange.

PR #364, opened 2023-07-28
show air in Block Statistic Search

EtlamGit:
When there are less non-air-Blocks on a depth layer than maximum shown Blocks, a blue background is used instead of gray to indicate air. This results in green: searched Blocks, gray: other Blocks, blue: air


EtlamGit:
grafik

PR #367, opened 2023-08-07
QRegExp -> QRegularExpression in most places

Kolcha:
QRegExp is deprecated since initial Qt 5 release

the only place that still uses QRegExp is search

### DIFF
--- a/minutor.cpp
+++ b/minutor.cpp
@@ -9,7 +9,7 @@
 #include <QtWidgets/QTreeWidget>
 #include <QProgressDialog>
 #include <QDir>
-#include <QRegExp>
+#include <QRegularExpression>
 #include <QtCore/QJsonDocument>
 #include <QtCore/QJsonArray>
 #include <QtCore/QJsonObject>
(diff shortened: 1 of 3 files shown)


EtlamGit:
As I understand it, `QRegularExpression` was recommended instead of `QRegExp` since Qt5.
But not deprecated. It is deprecated with Qt6 now.

I still struggle a bit with dropping Qt5.9 support completely.
Is there a way to not use [QRegularExpression::wildcardToRegularExpression](https://doc.qt.io/qt-5/qregularexpression.html#wildcardToRegularExpression-1) ?

Kolcha:
with Qt6 `QRegExp` is completely removed, not even deprecated. it is included only [Qt5Compat library](https://doc.qt.io/qt-6/qtcore5compat-module.html)

> Is there a way to not use [QRegularExpression::wildcardToRegularExpression](https://doc.qt.io/qt-5/qregularexpression.html#wildcardToRegularExpression-1) ?
good question... I see only the option to implement something "handcrafted", but this is maybe only simple if we need to handle simple cases (aka map "*" in pattern to ".*" in regex, and "?" to ".", which is maybe even irrelevant).
if some "supported wildcard specification" is here, I can look into this. for now I just replicated the same behavior I was able to "touch" and considered reasonable.

Kolcha:
dropped changes in search, left just a "safe" changes

PR #371, opened 2023-11-14
parse datapacks for their name

EtlamGit:
Parse all enabled datapacks for their real name (not the filename). Put all found datapacks in a list in WorldInfo, support zipped and unzipped storage. Can be used to query whether a specific datapack is used in that world.

Refactored Custom Dimension parsing into the same code.

Example usage:
Just call e.g. `WorldInfo::isDatapackEnabled("Chunklock")` to query if the datapack (in this case "Chunklock") is used.

### DIFF
--- a/worldinfo.cpp
+++ b/worldinfo.cpp
@@ -39,7 +39,11 @@ void WorldInfo::clear() {
 }
 
 
-bool WorldInfo::parseFolder(const QDir &path)
+// used to generate menu File->Open World
+// parse only basic stuff in the world folder:
+//  does it look like a Minecraft world
+//  does the world have a custom name
+bool WorldInfo::parseWorldFolder(const QDir &path)
 {
   clear();
   if (!path.exists("level.dat")) // no level.dat?  no world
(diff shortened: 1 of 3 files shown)


worldinfo.cpp
```diff
@@ -39,7 +39,11 @@ void WorldInfo::clear() {
 }
 
 
-bool WorldInfo::parseFolder(const QDir &path)
+// used to generate menu File->Open World
+// parse only basic stuff in the world folder:
+//  does it look like a Minecraft world
+//  does the world have a custom name
```
madmaxoft, line 45: This should have been a doxy-style comment in the header instead. Here it's buried.

worldinfo.cpp
```diff
@@ -88,74 +107,130 @@ bool WorldInfo::parseFolder(const QDir &path)
...
-  return true;
+
+bool WorldInfo::parsePackMcmeta(const QJsonDocument & json_doc, const QString path, bool enabled, bool zipped)
+{
+  QJsonObject json = json_doc.object();
+  if (json.contains("pack")) {
+    QJsonObject pack =  json.value("pack").toObject();
+    if (pack.contains("description")) {
+      DatapackInfo dp;
+      dp.name    = pack.value("description").toString();
```
madmaxoft, line 200: Is `description` the right field to consider an identifier? I'd guess that this is the user-facing description of the pack, rather than an identifier.

madmaxoft:
Does this work for server worlds, too? Are the folders the same?

EtlamGit:
Yes, in (downloaded) server worlds it is the same.

worldinfo.cpp
EtlamGit: You might be right here ... I have to think about it, and look into some more datapacks.
There is some structure defining the namespace, we should use that.

EtlamGit: https://minecraft.wiki/w/Data_pack#Folder_structure -> [show]

madmaxoft: There is structure, but not a single field that would serve as a unique identifier. Nothing we should be basing detection upon.

EtlamGit: The namespace of a mod has to be unique and can not be changed without breaking the mod.

EtlamGit: #372

PR #373, opened 2023-11-21
Fix playname query

EtlamGit:
Mojang has changed their API (deprecating the part we used in the past)

This PullRequest
1) changes the API entry point to get the player name to a variant that still works.
2) downloads the player skin texture and uses the face as icon in the jump to player menu

All player data is cached and only queried once.

### DIFF
--- a/minutor.cpp
+++ b/minutor.cpp
@@ -119,6 +119,10 @@ Minutor::Minutor()
   connect(this,    SIGNAL(worldLoaded(bool)),
           depth,   SLOT  (setEnabled(bool)));
 
+  // player cache request/response to Mojang API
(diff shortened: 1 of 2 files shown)


EtlamGit:
grafik

mrkite:
Beautiful.

EtlamGit:
You are very fast with merging today . . .
There is a compile issue with Qt 5.9.
I will add a PullRequest for that!

mrkite:
Yeah I saw that.  If it's easy to fix, by all means.

EtlamGit:
Yes, the fix was ready at the moment I realized that you merged already ;-)

PR #396, opened 2026-04-27
Qt6 compatibility patch 1

Kolcha:
- added missing includes

  some implicit dependencies allowed to build successfully, but with Qt6 these dependencies has changed. also it is a good practice to include what is used.

- don't use Qt module in some includes

  QAction is in QtGui in Qt6, but in QtWidgets in Qt5, just drop module name from include, use raw QAction.

- use `auto` for some iterator due to type mismatch

  looks like there was QMap -> QMultimap change without updating iterator type where it was created, but Qt5 was ok with this (a bit weird, but still). with Qt6 iterator types became incompatible (expected). since C++11 it is possible to use `auto`, and using it for iterator types is a common practice. project uses C++14, so using `auto` here is a good option.

- put C++17 compatibility stuff behind proper define guard

  project uses C++14 and has some stuff implemented that available in C++17. Qt6 requires C++17, so 2 implementations cause build issues. also it is a good practice to include compatibility stuff conditionally.

- adjusted .gitignore to handle modern Qt Creator layout

  Qt Creator creates `.qtcreator` directory to store project- specific information, and `build` for the build products.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ Makefile.Release
 *.pdb
 *.user
 *.pro.user.*
+
+/.qtcreator/
+/build/
--- a/clamp.h
+++ b/clamp.h
@@ -2,6 +2,8 @@
 #ifndef CLAMP_H_
 #define CLAMP_H_
 
+#if __cplusplus < 201703L
+
 namespace std {
 
 // will be in STD with C++17
@@ -12,4 +14,7 @@ const T& clamp( const T& value, const T& lower, const T& upper )
 }
 
 }
+
+#endif
+
 #endif  // CLAMP_H_
--- a/minutor.cpp
+++ b/minutor.cpp
@@ -1,6 +1,6 @@
 /** Copyright (c) 2013, Sean Kasun */
 #include <QtWidgets/QVBoxLayout>
-#include <QtWidgets/QAction>
+#include <QAction>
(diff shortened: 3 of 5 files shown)


Kolcha:
Qt6 migration plan is to use [Qt5 compatibility module](https://doc.qt.io/qt-6/qtcore5-index.html), so no "intrusive changes" assumed (but still some changes are required). both Qt5 and Qt6 will be supported after migration. changes will be introduced gradually.

EtlamGit:
All your point sound very useful, will try to look into the changes when I find some time.

EtlamGit (approved):

EtlamGit:
Just some comments to keep you motivated:
* I still struggle with the Qt6 license terms and it feels wrong for me to use Qt6 in the current state.
* Nevertheless making the code compatible with Qt5 and Qt6 is perfectly fine and good coding practice.
* My spare time is extremely limited and the time budget for Minutor is mostly consumed by compatibility fixes and version updates. Therefore it might take many weeks, sometimes even until next big vacation period that I will find time to take a look into topics I do not like that much.

So: Do not get demotivated when you do not receive any feedback. Just place a ping in the PR every month to poke me. Also try to make smaller PRs with good explanation, because that reduces the review time. It seems that you tried that tactic this time, as this PR is a perfect example for that ;-)
I really appreciate your effort to improve code quality!

PR #397, opened 2026-04-28
Qt6 compatibility patch 2

Kolcha:
this patch resolves compatibility issues with QString::split()

- QRegExp -> QRegularExpression migration (in split() cases only)

  QString::split() accepting QRegExp is not available in Qt 6 (also deprecated in Qt 5), so such cases must be migrated. there was only single such case, and it is 1:1 migration.

- QString::SplitBehavior -> Qt::SplitBehavior migration

  QString::SplitBehavior is obsolete since Qt 5.14, and was completely removed in Qt 6. use new values in the codebase and introduce compatibility layer for the older Qt versions.

### DIFF
--- a/compat.h
+++ b/compat.h
@@ -1,0 +1,19 @@
+// Copyright (c) 2026, Nick Korotysh
+#pragma once
+
+/*
+ * This file contains various compatibility stuff
+ * to support builds with both Qt 5 and Qt 6.
+ *
+ * It should be dropped with Qt 5 build support.
+ */
+
+#if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
+#include <QString>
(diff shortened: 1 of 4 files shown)


Kolcha:
these changes are completely independent from the changes in the previous PR (#396) and can be merged in any order. this PR is also "micro PR" (as the previous one) - very few modified files with few changed lines. likely only one more PR will be created later to introduces the final changes (also minimal) to enable Qt6 builds.

EtlamGit:
Just as reference:
https://doc.qt.io/archives/qt-5.15/qstring-obsolete.html
https://stackoverflow.com/questions/74416992/error-skipemptyparts-is-not-a-member-of-qt

PR #398, opened 2026-04-28
Qt6 compatibility patch 3

Kolcha:
- conditionally depend on core5compat module

  core5compat contains QRegExp and QStringRef classes that are not available in Qt 6 by default.

- replaced QString::midRef() with QString::mid()

  QString::midRef() is not available in Qt 6 due to QStringRef removal. replacing it with QString::mid() is not the best option as this introduces memory allocation, but the easiest one.

  QStringView can't cover all cases as it is limited in Qt 5. but this can be addressed later with full Qt 6 migration.

### DIFF
--- a/minutor.pro
+++ b/minutor.pro
@@ -2,6 +2,7 @@ TEMPLATE = app
 TARGET = minutor
 CONFIG += c++14
 QT += widgets network concurrent
+greaterThan(QT_MAJOR_VERSION, 5) QT += core5compat
 QMAKE_INFO_PLIST = minutor.plist
 unix:LIBS += -lz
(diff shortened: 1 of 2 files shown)


Kolcha:
this is the last PR required to enable Qt 6 builds. it also contains independent changes as the previous ones (#396, #397) and can be merged in any order. this PR even smaller than the previous ones.

after merging all these 3 PRs Qt 6 builds can be enabled.

EtlamGit:
Just as reference:
https://doc.qt.io/qt-6/qtcore-changes-qt6.html#the-qstringref-class

EtlamGit (approved):

PR #399, opened 2026-06-27
Add UI setting for zoom follows cursor behavior

terbin:
## Summary

Minutor already zooms toward the mouse cursor on wheel zoom, but the setting that controls it (`zoomFollowsCursor`) has only ever existed as a hidden `QSettings` value, with no way to change it short of editing settings by hand. The commit that added the behavior, #295 (`36a9d02`), explicitly anticipated this follow-up: "a future commit could expose this setting in the GUI, making it possible to disable the new zoom behavior." This PR adds that checkbox to the Settings dialog.

My use case is comparing the same place in two worlds from different points in time, to spot what changed between them. That comparison works best when the view stays centered on the same spot as you zoom in for a closer look. The cursor-following default instead re-centers on wherever the mouse is, which drifts the framing and pushes the two views out of alignment. Unchecking the box restores center-anchored zoom and holds the framing steady, so the two stay lined up.

## What changed

- A "Zoom toward mouse cursor" checkbox in the Settings dialog, wired to the existing `zoomFollowsCursor` key the same way the other checkboxes are (load, default, save).
- It sits in the existing "Mouse Wheel" group, beside the related "Depth Slider" and "Zoom Out" controls, since the setting only affects wheel zoom, not keyboard zoom.

## Behavior

- Checked (default): wheel zoom keeps the point under the cursor fixed.
- Unchecked: wheel zoom keeps the view center fixed.
- Read live on each wheel event, so it takes effect immediately, with no restart needed.
- The default is unchanged, so existing users see no difference; it only makes the option reachable.

## Screenshot

Settings dialog with the new checkbox

## Notes

- No change to the `zoomFollowsCursor` key or its semantics; this only adds the missing UI.
- The slot doesn't emit `settingsUpdated()`: the value is read live at zoom time, so no redraw is needed.
- Builds clean with `qmake && make` (Qt 5).
- The `settings.ui` diff is mostly whitespace. Adding the checkbox below the existing row meant wrapping that row in a vertical layout, which re-indents the surrounding markup. The real change is about 14 lines.

### DIFF
--- a/settings.cpp
+++ b/settings.cpp
@@ -58,6 +58,9 @@ Settings::Settings(QWidget *parent) : QDialog(parent) {
   connect(m_ui.radioButton_zoom_alt,   &QRadioButton::toggled,
           this,                        &Settings::toggleModifier4ZoomOut);
 
+  connect(m_ui.checkBox_ZoomFollowsCursor, SIGNAL(toggled(bool)),
+          this, SLOT(toggleZoomFollowsCursor(bool)));
+
   // Load the settings:
   QSettings info;
   auto useDefault = info.value("usedefault", true).toBool();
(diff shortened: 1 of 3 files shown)


terbin:
I know you are are busy, no hurries on reviewing this please. I just wanted to put it here :)

EtlamGit:
Looks fine to me. Also nice prepared PR description.
And I'm quite surprised that the setting was already present for 4 years now...

I understand your use case.
When I came across that problem (want to zoom without pan) I used keyboard zoom (PageUp, Page Down) to get the desired result. With that you have both behaviors without changing a setting.

But of course, having that setting is fine.

terbin:
Thank you :)
Reaching for the Page Up, Page Down buttons works in most cases. But sometimes working with the mouse is easier, if you are in a mouse-workflow anyway. Else you have to move your hand to keyboard.

